PR #5, opened 2021-02-15
[SIE-46] create Loader component

alecherryy:


alecherryy:
**Note:** This is just a static component while we wait to gain more details on the implementation.

quanganhhoang (approved):
LGTM!

PR #9, opened 2021-02-16
Feature/sie 17 create grid layout

denglish7:
Created grid layout

### DIFF
--- a/src/stories/layouts/Grid/Grid.js
+++ b/src/stories/layouts/Grid/Grid.js
@@ -7,7 +7,7 @@ import PropTypes from 'prop-types';
  * Component for grid layout element.
  *
  * @param {node} children of the component
- * @param {string} numColumns number of columns in layout.
+ * @param {number} numColumns number of columns in layout.
  * @return {object} (
  *   <Grid numColumns={numColumns}>
  *      {children}
@@ -16,8 +16,10 @@ import PropTypes from 'prop-types';
  */
 
 export const Grid = ({ children, numColumns }) => {
+    const modifierClass = numColumns ? `grid--${numColumns}-col` : '';
+
     return (
-        <div className={['grid', `${numColumns}`].join(' ').trim()}>
+        <div className={['grid', modifierClass].join(' ').trim()}>
             {children}
         </div>
     );
@@ -31,7 +33,7 @@ Grid.propTypes = {
     /**
    * Grid's number of columns
    */
-    numColumns: PropTypes.string,
+    numColumns: PropTypes.number,
 };
 
 Grid.defaultProps = {
--- a/src/stories/layouts/Grid/Grid.js
+++ b/src/stories/layouts/Grid/Grid.js
@@ -1,0 +1,39 @@
+import './styles.scss';
+
+import React from 'react';
+import PropTypes from 'prop-types';
+
+/**
+ * Component for grid layout element.
+ *
+ * @param {node} children of the component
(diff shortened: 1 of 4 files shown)


src/stories/layouts/Grid/Grid.stories.js
```diff
@@ -0,0 +1,68 @@
...
+ */
+export default {
+    title: 'Example/Grid',
+    layout: Grid,
+};
+
+const Template = (args) => (
+    <Grid numColumns={args.numColumns}>
+        <div className="grid__item">
+            <h1>Test</h1>
```
alecherryy, line 17: I would updated the content of the sample grid item to be an `h3` and include a paragraph with sample content, like so:
```
<div className="grid__item">
    <h3>Test</h3>
    <p>Lorem ipsum dolor sit amet, aliquip pertinax intellegebat vel ad. 
    Quo diceret comprehensam id, mazim labore instructior et mea.</p>
</div>
```

src/stories/layouts/Grid/styles.scss
```diff
@@ -0,0 +1,56 @@
...
+.grid--four {
+    @extend %shared;
+    grid-template-columns: repeat(4, 1fr);
+}
+
+.grid__item {
+    display: flex;
+    align-items: center;
+    justify-content: center;
+}
```
alecherryy, line 52: Remove this. We don't want to make assumptions on the display of the grid items.

src/stories/layouts/Grid/Grid.js
```diff
@@ -0,0 +1,39 @@
...
+ * @return {object} (
+ *   <Grid numColumns={numColumns}>
+ *      {children}
+ *    </Grid>
+ * )
+ */
+
+export const Grid = ({ children, numColumns }) => {
+    return (
+        <div className={['grid', `${numColumns}`].join(' ').trim()}>
```
alecherryy, line 20: I like the idea of using the number of columns; I would just update this to `grid--${numColumns}-col` and change the `numColumns` param to a number, which would return the following modifier class: `grid--3-col`.  That way we don't need to spell out numbers which is more likely to lead to typos and then bugs. Example of update:
```
const modifierClass = numColumns ? `grid--${numColumns}-col` : '';
    return (
        <div className={['grid', modifierClass].join(' ').trim()}>
            {children}
        </div>
    );
```

src/stories/layouts/Grid/styles.scss
```diff
@@ -0,0 +1,56 @@
...
+    }
+
+    @include bp-desktop {
+        grid-template-columns: repeat(4, 1fr);
+        grid-column-gap: $spacing-3;
+    }
+}
+
+.grid--one {
+    @extend %shared;
```
alecherryy, line 29: You don't need this. Those styles are already present in the `.grid` class.

alecherryy (changes requested):
@denglish7 I left a couple of small suggestions/changes; however, I noticed your grid is not responsive. You might want to double check that!

denglish7:
> @denglish7 I left a couple of small suggestions/changes; however, I noticed your grid is not responsive. You might want to double check that!

denglish7:
> > @denglish7 I left a couple of small suggestions/changes; however, I noticed your grid is not responsive. You might want to double check that!

Thank you! I have made the requested changes and the grid is now responsive with 1-4 columns

alecherryy (approved):

PR #11, opened 2021-02-18
[SIE 14] Create Branding component

fernandowinfield:
- Created a Branding component that can be displayed at the top of pages.

- The Branding component includes a logo and a name of the brand. For now I have used the polygon from Sketch as the placeholder for the logo and the name of the brand is passed as a prop to the component. The Branding altogether is a clickable <a> element that navigates to home (`/`).

- Jira task: https://cs6510.atlassian.net/browse/SIE-14


src/stories/components/Branding/styles.scss
```diff
@@ -0,0 +1,14 @@
+@import '../../../scss/variables.scss';
+
+// @file
+// Branding
+
+.branding {
+  font-size: $font-size-xl;
+  font-weight: bold;
```
fernandowinfield, line 8: I couldn't find a variable for bold font-weight in the presets that Alessia added so I added this here. But if it would be better to factor that out for later reuse I can try to do that.

src/stories/components/Branding/styles.scss
```diff
@@ -0,0 +1,14 @@
...
+  font-family: $header-font-family;
+}
+
+.logo {
+  margin-right: 5px;
```
fernandowinfield, line 13: I decided to use 'pixels' here instead of 'rem' for the separation between the logo and the brand name because I thought the branding should remain consistent throughout the website. But if you think 'rem' is more appropriate here I can easily change it using the function provided by Alessia.

src/stories/components/Branding/styles.scss
```diff
@@ -0,0 +1,14 @@
...
+// @file
+// Branding
+
+.branding {
+  font-size: $font-size-xl;
+  font-weight: bold;
+  font-family: $header-font-family;
+}
+
+.logo {
```
fernandowinfield, line 12: In terms of the alignment of the logo, right now it is not perfectly aligned vertically inside the anchor element. In other words, this makes the logo appear higher compared to how it shows on the Sketch design. I wonder if we want the logo to be perfectly aligned vertically. If this is the case I might need help from someone who is experienced in little styling details like this one.

src/stories/components/Branding/Branding.js
```diff
@@ -0,0 +1,50 @@
...
+export const Branding = ({ modifierClasses, logo, text }) => {
+    return (
+        <a
+            href="*/"
+            className={['branding', `${modifierClasses}`].join(' ').trim()}
+        >
+            <img src={Logo}
+                alt={`${text} website logo`}
+                className="logo"/>
+            {text}
```
fernandowinfield, line 28: I decided to just use the text inside the anchor tag (<a><a/>) as the brand name. I wonder if it would be better to use a header tag instead (such as <h1><h1/>)...

src/stories/components/Branding/styles.scss
stenizwahyudi: I think it should be fine if the one you need is not defined in the global variable. If not, it should be a very easy fix.

src/stories/components/Branding/styles.scss
stenizwahyudi: Not really sure about this, but I remember we talked about using rem as it is more reliable than px, so I assume it is better to use margin-right: rem(5px);

src/stories/components/Branding/styles.scss
stenizwahyudi: I tried to add some styling to the .logo and it fixes your issue. You can try this:
.logo {
   margin-right: rem(5px);
   position: absolute;
   top: rem(3px);
}

src/stories/components/Branding/Branding.js
stenizwahyudi: Not sure if one is better than the other, but if either one is actually better, I'm sure it should be a very easy fix.

src/stories/components/Branding/Branding.js
```diff
@@ -0,0 +1,52 @@
...
+
+Branding.propTypes = {
+    /**
+   * Branding's modifier classes
+   */
+    modifierClasses: PropTypes.string,
+    /**
+   * Branding's logo URL
+   */
+    logo: PropTypes.string.isRequired,
```
stenizwahyudi, line 43: Should logo also be declared as component parameter above? (see line 11-16)

src/stories/components/Branding/Branding.stories.js
```diff
@@ -0,0 +1,20 @@
...
+    component: Branding,
+};
+
+const Template = (args) => <Branding {...args} />;
+
+// Default branding
+export const Default = Template.bind({});
+Default.args = {
+    modifierClasses: '',
+    url: 'src/images/Polygon.png',
```
stenizwahyudi, line 18: Maybe change url to logo to be consistent.

stenizwahyudi (changes requested):
I finished reviewing your pull request for now and suggested some changes. However, I highly recommend to ask for second opinion before actually making any changes.

src/stories/components/Branding/styles.scss
alecherryy: @fernandowinfield no variable for font weights. You can use `bold` or `700` to achieve that!

src/stories/components/Branding/styles.scss
alecherryy: @fernandowinfield you can do the following:
```
.branding {
  align-items: center;
  display: inline-flex;
}
```

src/stories/components/Branding/Branding.js
```diff
@@ -0,0 +1,52 @@
...
+    // TODO: replace below <a> with <Link> once routing is set up OR modify
+    //       href of below <a> once page navigation is established.
+    return (
+        <a
+            href="/"
+            className={['branding', `${modifierClasses}`].join(' ').trim()}
+        >
+            <img src={Logo}
+                alt={`${text} website logo`}
+                className="logo"/>
```
alecherryy, line 29: @fernandowinfield `className` should be `branding__logo`, following the BEM naming convention.

src/stories/components/Branding/Branding.js
alecherryy: You cannot nest block elements (`<h1>`, `<p>`, etc) inside inline elements (like `<a>`, `<span>`, etc). This is okay for now!

src/stories/components/Branding/Branding.js
```diff
@@ -0,0 +1,52 @@
+import './styles.scss';
+
+import React from 'react';
+import PropTypes from 'prop-types';
+// TODO: replace the import below with the actual logo once it's chosen
+import Logo from '../../../images/Polygon.png';
```
alecherryy, line 6: @fernandowinfield let's use an SVG file instead of PNG, that way it won't lose quality. You can export it as an SVG directly from Sketch.

src/stories/components/Branding/styles.scss
alecherryy: I think this okay for now.

alecherryy (changes requested):

src/stories/components/Branding/Branding.js
fernandowinfield: Whoops, I forgot to update that part. We can actually get rid of that parameter if we just load a predefined logo directly from the component. Good catch!

fernandowinfield:
- Removed unnecessary component parameter (`logo`)
- Replaced .png logo with a .svg
- Fixed className of logo to follow BEM convention
- Added recommended styling to improve the vertical alignment of the branding elements (logo and name)

alecherryy (approved):

stenizwahyudi (approved):

PR #14, opened 2021-02-19
Feature/sie 15 card

cmutuc25:


src/stories/components/Card/Card.js
```diff
@@ -0,0 +1,85 @@
...
+     */
+    modifierClasses: PropTypes.string,
+    /**
+     * Card's title
+     */
+    title: PropTypes.string.isRequired,
+    /**
+     * Card's body
+     */
+    body: PropTypes.string.isRequired,
```
stenizwahyudi, line 68: I saw that the short description (in your case the body param) is optional in Jira. Do you think we should remove the "isRequired" for body?

src/stories/components/Card/styles.scss
```diff
@@ -0,0 +1,80 @@
...
+
+    .card__button {
+        &:hover,
+        &:focus,
+        &:active,
+        &:visited {
+            background-color: $black;
+            border-radius:0;
+            border-bottom-right-radius: 50px;
+            border-top-left-radius: 50px;
```
stenizwahyudi, line 65: I believe we can use $spacing-5 for both line 64 and 65.

src/stories/components/Card/styles.scss
```diff
@@ -0,0 +1,80 @@
+@import '../../../scss/variables.scss';
+
+// @file
+// Card
+
+.card
+{
+    border-radius: 5px;
+    box-shadow: 0 4px 8px 0 rgba(0,0,0,0.2);
```
stenizwahyudi, line 9: Not 100% sure but maybe convert the px to rem function.
i.e. border-radius: rem(5px);

src/stories/components/Card/styles.scss
cmutuc25: This css is written to prevent the changing of the button. I think this should stay as `50px` to be consistent with the button's css, which is also set at `50px`

src/stories/components/Card/styles.scss
cmutuc25: This should stay as is. We don't want the border shadow effect to change sizes when if the font size changes

src/stories/components/Card/Card.js
cmutuc25: done

src/stories/components/Card/styles.scss
stenizwahyudi: I see, looks good to me!

src/stories/components/Card/Card.js
stenizwahyudi: Thank you!

stenizwahyudi (approved):

PR #16, opened 2021-02-22
[SIE-19] Form Component

harmanwho:
Feel free to comment if any changes are needed.
There are some warnings in the console, which I wasn't able to recognize the reason for. Also, the forms are not optimized for dark/green background as FormItems don't have a dark theme.


src/stories/components/Form/Form.js
```diff
@@ -0,0 +1,211 @@
...
+
+/**
+ * Render <Login /> HTML
+ * @return {object} (
+ *   <Login />
+ * )
+ */
+const Login = () => (
+    <React.Fragment>
+        <div className="form__item">
```
alecherryy, line 147: You don't need these wrapper `div`'s. Form items already have a wrapper around them, if you need it for styling, feel free to update margins, spacing or whatever else here: https://github.com/neu-self-image-experiments/sie-firebase/blob/master/src/stories/components/FormItem/styles.scss

src/stories/components/Form/styles.scss
```diff
@@ -0,0 +1,20 @@
+@import '../../../scss/variables.scss';
+
+// @file
+// Form
+
+.form__container{
+    display: flex;
```
alecherryy, line 7: Any reason why you choose to use `display: flex`? If it's just because of the `Submit` button, I can show you how to align it right without having to call flexbox.

src/stories/components/Form/styles.scss
```diff
@@ -0,0 +1,20 @@
...
+    justify-content: flex-end;
+    padding: 0 0 $spacing-1;
+}
+
+.form__item {
+    flex-basis: 100%;
+    margin: 0 0 $spacing-2;
+}
+
+.form__button {
```
alecherryy, line 18: Follow up on the comment above: I would remove everything that is flex-related and simply add `text-align: right` here.

alecherryy (changes requested):
> Feel free to comment if any changes are needed.
> There are some warnings in the console, which I wasn't able to recognize the reason for. Also, the forms are not optimized for dark/green background as FormItems don't have a dark theme.

@harmanwho `FormItem` has a modifier class called `.form-item--light` which applies light styling to input fields and labels. If you want to implement it, I would add a boolean `@param` to Form (something called `isDarkTheme` and have each `FormItem` check its value and dynamically add the modifier class to themselves if needed. Example:

```
const Login = ({ isDarkTheme} ) => (
    <React.Fragment>
        <FormItem
            modifierClasses={isDarkTheme && 'form-item--light'}
            placeholder="Email"
            type="email"
            showLabel
        />
        <FormItem
            modifierClasses={isDarkTheme && 'form-item--light'}
            placeholder="Password"
            type="password"
            showLabel
        />
    </React.Fragment>
);
```
Parent component:
```
export const Form = (
    { isDarkTheme, formItems, type, buttonText, handleSubmit },
) => {
    // dynamically render Form Item Input
    function returnForm(type) {
        // switch statement to check value of type
        switch (type) {
        case 'login':
            return <Login isDarkTheme={isDarkTheme} />;
        case 'signup':
            return <SignUp isDarkTheme={isDarkTheme} />;
        default:
            return <CustomForm isDarkTheme={isDarkTheme} elements={formItems} />;
        }
    }
    return (
        // your markup here
    );
};
```

src/stories/components/Form/styles.scss
harmanwho: You are right, I didn't need to use display flex for the submit button, your solution fixed it

src/stories/components/Form/Form.js
harmanwho: I need the div to stretch the form element to the full-width of the container using `flex-basis: 100%`, do you suggest I make this change in the `<FormItem />` component itself?

harmanwho:
> > Feel free to comment if any changes are needed.
> > There are some warnings in the console, which I wasn't able to recognize the reason for. Also, the forms are not optimized for dark/green background as FormItems don't have a dark theme.
> 
> @harmanwho `FormItem` has a modifier class called `.form-item--light` which applies light styling to input fields and labels. If you want to implement it, I would add a boolean `@param` to Form (something called `isDarkTheme` and have each `FormItem` check its value and dynamically add the modifier class to themselves if needed. Example:
> 
> ```
> const Login = ({ isDarkTheme} ) => (
>     <React.Fragment>
>         <FormItem
>             modifierClasses={isDarkTheme && 'form-item--light'}
>             placeholder="Email"
>             type="email"
>             showLabel
>         />
>         <FormItem
>             modifierClasses={isDarkTheme && 'form-item--light'}
>             placeholder="Password"
>             type="password"
>             showLabel
>         />
>     </React.Fragment>
> );
> ```
> 
> Parent component:
> 
> ```
> export const Form = (
>     { isDarkTheme, formItems, type, buttonText, handleSubmit },
> ) => {
>     // dynamically render Form Item Input
>     function returnForm(type) {
>         // switch statement to check value of type
>         switch (type) {
>         case 'login':
>             return <Login isDarkTheme={isDarkTheme} />;
>         case 'signup':
>             return <SignUp isDarkTheme={isDarkTheme} />;
>         default:
>             return <CustomForm isDarkTheme={isDarkTheme} elements={formItems} />;
>         }
>     }
>     return (
>         // your markup here
>     );
> };
> ```

Oh cool, I didn't thoroughly check the component - my bad

src/stories/components/Form/Form.js
harmanwho: Figured it out, too many div wrappers were causing the issue

alecherryy (approved):

PR #17, opened 2021-02-23
Created Teaser Component

stenizwahyudi:
Description
As a front end developer, I want to create a Teaser component; the entire component should be linked and it should include the following fields:
1. URL
2. Title
3. Small Text

The teaser comes in three styles:
1. With a Dark Teal Line
2. With a Teal Line
3. With a Gray Line and Gray Line (when the link is disabled)

Each Teaser has a light gray bottom border and an arrow icon which can be found on Google Material Design. Icon files should be named according to the following convention: icon-{descriptive-name}.svg

Sketch: https://www.sketch.com/s/3e6b9c91-118c-471e-98ad-8676d41d1435/a/zxyGnLW


src/stories/components/Teaser/styles.scss
```diff
@@ -0,0 +1,56 @@
+@import '../../../scss/variables.scss';
+
+// @file
+// Teaser
+
+.teaser {
+    width: 100%;
+    border: solid lightgrey;
```
stenizwahyudi, line 8: The color from the global variable doesn't match the sketch, so I used lightgrey instead.

src/stories/components/Teaser/styles.scss
```diff
@@ -0,0 +1,56 @@
...
+    }
+}
+
+.teaser--disabled {
+    color: grey;
+    .arrow__logo {
+        visibility: hidden;
+    }
+    .line {
+        border: solid grey;
```
stenizwahyudi, line 54: The color from the global variable doesn't match the sketch, so I used grey instead.

src/stories/components/Teaser/Teaser.stories.js
```diff
@@ -0,0 +1,39 @@
...
+    url: '#',
+    title: 'Title with a Teal Line',
+    text: 'Teal line',
+};
+
+// Disabled teaser
+export const Disabled = Template.bind({});
+Disabled.args = {
+    modifierClasses: 'teaser--disabled',
+    url: '#',
```
cmutuc25, line 36: The disabled teaser should probably not link somewhere. Maybe disable the link and keep the url to pass in?

cmutuc25 (commented):
Left a comment about the disabled type

src/stories/components/Teaser/Teaser.stories.js
stenizwahyudi: That is a very good observation.

stenizwahyudi: I disabled the link redirection in the styles.scss (see line 49)

src/stories/components/Teaser/styles.scss
alecherryy: I'll double check this. The repo might be out of date or the Sketch file is wrong.

alecherryy (commented):
@stenizwahyudi all looks good, except for the `className`'s. Don't forget to use the BEM convention; this way you don't have to nest elements to bind the styles. 

I'm gonna update the component to include a fun hover animation.

src/stories/components/Teaser/Teaser.js
```diff
@@ -0,0 +1,63 @@
...
+    return (
+        <a
+            href={url}
+            className={['teaser', `${modifierClasses}`].join(' ').trim()}
+        >
+            <img src={Logo}
+                alt={`${text} arrow logo`}
+                className="arrow__logo"
+            />
+            <div className='line'></div>
```
alecherryy, line 31: Rather than adding extra markup, you can use `pseudo-elements` to add decorative elements.

stenizwahyudi: Aha, I just knew about this. Done, thank you!

alecherryy (approved):

PR #18, opened 2021-02-23
[SIE-16] create Header component

alecherryy:


denglish7 (approved):
Looks great! I like moving Constrain component to the layout folder, makes more sense to me.

PR #19, opened 2021-02-23
[SIE-16] remove Regions folder and use layouts instead

alecherryy:


alecherryy:
@denglish7 I decided to move `Header` into layout as well since we won't have enough regions and the content of he header will change between screens.

denglish7:
> @denglish7 I decided to move `Header` into layout as well since we won't have enough regions and the content of he header will change between screens.

Makes sense!

denglish7 (approved):

PR #20, opened 2021-02-23
Feature/sie 21 data cards

denglish7:
Created data card component


src/stories/components/DataCard/DataCard.js
```diff
@@ -0,0 +1,61 @@
...
+ * )
+ */
+export const DataCard = (
+    { modifierClasses, title, description, children },
+) => {
+    return (
+        <div className={['data-card', `${modifierClasses}`].join(' ').trim()}>
+            <div className="data-card__container">
+                <div className="data-card__details">
+                    <p className="data-card__title">{title}</p>
```
alecherryy, line 29: @denglish7 Update this to `<h6>` and you won't need to apply special styles. I realized there's an error in the global styles, sorry! Can you replace this [line](https://github.com/neu-self-image-experiments/sie-firebase/blob/4ab821ae0073ad1f40bf3170842420fbdb49e9f2/src/scss/_base.scss#L23) with `font-family: $header-font-family;`. Thank you!

src/stories/components/DataCard/styles.scss
```diff
@@ -0,0 +1,45 @@
+@import '../../../scss/variables.scss';
+
+// @file
+// Data Card
+
+.data-card {
+    border-radius: 5px;
```
alecherryy, line 7: Nitpick-y design thing, can you change this to be `20px`? Thanks!

src/stories/components/DataCard/styles.scss
```diff
@@ -0,0 +1,45 @@
...
+.data-card__title {
+    font-weight: 700;
+}
+
+.data-card__details {
+    margin-bottom: $spacing-5;
+    p {
+        margin-bottom: $spacing-1;
+        font-size: $font-size-md;
+        font-family: $header-font-family;
```
alecherryy, line 29: No need for these styles. The small text is using the default body font family.

src/stories/components/DataCard/styles.scss
```diff
@@ -0,0 +1,45 @@
...
+        line-height: $line-height-tight;
```
alecherryy, line 30: You can apply this directly to the `.card` class and it will automatically apply it to the paragraph.

src/stories/components/DataCard/styles.scss
```diff
@@ -0,0 +1,45 @@
...
+    }
+}
+
+.data-card__teal {
+    background-image: $gradient-teal;  
+}
+
+.data-card__teal-dark {
+    background-image: $gradient-teal-dark;
+    color: $white;
```
alecherryy, line 40: After you update the title to use an `<h6>`, you are going to have to override the `.data-card__title` color here.

alecherryy (changes requested):

alecherryy:
@denglish7 really small changes, 2 of which are because I made a mistake in the global styles. Thank you!

src/stories/components/DataCard/styles.scss
denglish7: Was wondering if this should be a certain value, thanks!

denglish7:
> @denglish7 really small changes, 2 of which are because I made a mistake in the global styles. Thank you!

Thank you! Updating now

alecherryy (approved):

PR #25, opened 2021-02-25
SIE-40 created CRUD-user and updated firebase.js and index.html

dustinlo:
Here is the user crud skeleton for now. Please review. Thanks!

### DIFF
--- a/public/index.html
+++ b/public/index.html
@@ -41,11 +41,11 @@
     -->
 
     <!-- The core Firebase JS SDK is always required and must be listed first -->
-    <script src="/__/firebase/8.2.6/firebase-app.js"></script>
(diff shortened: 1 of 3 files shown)


theodore3131 (approved):

quanganhhoang (changes requested):
Please remove `console.log` and have proper & consistent returns for your functions. I see some functions return true/false and others just silently do nothing.

dustinlo:
> Please remove `console.log` and have proper & consistent returns for your functions. I see some functions return true/false and others just silently do nothing.

Hello, I updated it according to ExperimentServices revised by Hao. Please re-review it again. Thanks!

quanganhhoang (approved):

PR #27, opened 2021-02-25
add self image CRUD service

theodore3131:
- defined basic CRUD operations in a separate selfImageExperimentService.js

### DIFF
--- a/src/firebase/CRUDServices/selfImageExperimentService.js
+++ b/src/firebase/CRUDServices/selfImageExperimentService.js
@@ -22,7 +22,7 @@ export default class SelfImageExperimentService {
     /**
      * Create self image experiment in database.
      * @param {String} experimentId id of the new self image experiment.
-     * @param {SelfImageExperiment} selfImageExperiment 
+     * @param {SelfImageExperiment} selfImageExperiment
      * self image experiment object.
      * @return {Boolean} succeed or not.
      */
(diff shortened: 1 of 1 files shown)


quanganhhoang (changes requested):
could you please remove `console.log` lines? They should not be in production code.

theodore3131:
> could you please remove `console.log` lines? They should not be in production code.

Sure, thanks, will update in a sec

dustinlo (approved):
All returns are consistent and there is no console.log no more.

src/firebase/CRUDServices/selfImageExperimentService.js
```diff
@@ -0,0 +1,116 @@
...
+        if (snapshot.empty) {
+          return;
+        }
+
+        snapshot.forEach( (doc) => {
+          // TODO: add logic here
+        });
+      } catch (err) {
+        // TODO: add logic here
+      }
```
quanganhhoang, line 57: shouldn't this return SelfImageExperiment[] per documentation? Both try and catch blocks are missing return statements

src/firebase/CRUDServices/selfImageExperimentService.js
```diff
@@ -0,0 +1,116 @@
...
+          .doc(experimentId);
+        const doc = await experimentRef.get();
+        if (!doc.exists) {
+          // TODO: add logic here
+        } else {
+          // TODO: add logic here
+        }
+        return doc;
+      } catch (err) {
+        // TODO: add logic here
```
quanganhhoang, line 78: branch missing here? If there is an error we are not returning anything

quanganhhoang (changes requested):

src/firebase/CRUDServices/selfImageExperimentService.js
theodore3131: Updated for review.

quanganhhoang (approved):

PR #28, opened 2021-02-25
sie-41-revice-console-log/error-handling

bleachalon:
1. Remove all console.log and console.error as requested.
2. Leave the error handling as return false and empty object/array now. Will update after decide which error handling method to use.


dustinlo (approved):
Look good to me. This is actually a skeleton of our crud so the logic can be updated later.

theodore3131 (approved):
lgtm

PR #30, opened 2021-02-28
[SIE-35] create Login Page

alecherryy:


src/stories/pages/Login/Login.js
```diff
@@ -0,0 +1,121 @@
...
+export const Login = ({ isDarkTheme }) => {
+  const [error, setError] = useState(false);
+  const [username, setUsername] = useState('');
+  const [password, setPassword] = useState('');
+
+
+  const getUser = (e) => {
+    e.preventDefault();
+
+    const userService = new UserServices();
```
bleachalon, line 34: Line 34 should be _const userService = UserServices.getInstance();_ that way it will trigger the singleton creation process and assign memory to object only if it does not exists, which prevents the program to repeatedly assign and trash memory every time service gets called. Everything else looks good! Thanks for being proactive.

bleachalon (commented):
Just one little advice. Everything else looks fine.

src/stories/pages/Login/Login.js
alecherryy: @bleachalon can you take a look at this branch? Calling `getInstance()` returns `undefined`.

PR #31, opened 2021-02-28
[SIE] add Footer and restructure Form to allow children

alecherryy:


alecherryy (commented):
Sorry about the huge amount of changes. Please, double check the `<Form />` since I had to re-structure the entire component. The reason for the changes is to allow `<FormItem />` to have their on `OnChange` parameter which we can use across the site to set the state. 

The rest are small changes and indentation adjustments.

There is no ticket for this.

denglish7 (approved):
Everything looks good, including the restructured form. Nice work!

PR #32, opened 2021-02-28
[SIE-72] dynamically set Firebase config from .env file

alecherryy:


bleachalon (approved):
Thank you for solving this!

PR #33, opened 2021-02-28
add check condition when instance undefined

bleachalon:
Fix the issue of getInstance() return undefined after change == to ===.
Change instance name to be more semantic.


src/firebase/CRUDServices/experimentServices.js
```diff
@@ -7,17 +7,18 @@ export default class ExperimentServices {
...
     /**
      * creating singleton instance.
      * @return {ExperimentServices} instance
      */
     static getInstance = () => {
-      if (self.sieInstance === null) {
-        self.sieInstance = new ExperimentServices();
+      if (self.experimentServiceInstance === null ||
+          self.experimentServiceInstance === undefined) {
+        self.experimentServiceInstance = new ExperimentServices();
```
quanganhhoang, line 19: can we do `if (!self.experimentServiceInstance)` here? I'm just curious

https://stackoverflow.com/questions/5515310/is-there-a-standard-function-to-check-for-null-undefined-or-blank-variables-in

alecherryy: Nit-picky thing: the file names do not match the class names. I.e. `experimentService` vs `ExperimentService`.

alecherryy:
@bleachalon Nit-picky thing: the file names do not match the class names. I.e. `experimentService.js` vs `ExperimentService`. They should both be capitalized.

alecherryy (approved):
Thank you for working on this. If you could update those two things, it'd be great!

PR #34, opened 2021-03-01
[SIE-35] add Login page

alecherryy:


src/stories/pages/Login/Login.js
```diff
@@ -0,0 +1,134 @@
...
+export const Login = ({ isDarkTheme }) => {
+  const [error, setError] = useState(false);
+  const [username, setUsername] = useState('');
+  const [password, setPassword] = useState('');
+
+  const getUser = (e) => {
+    e.preventDefault();
+
+    const service = UserServices.getInstance();
+    console.log(service.getUsers()); // eslint-disable-line no-console
```
cmutuc25, line 35: Is there a reason for this? Shouldnt we not be using `console.log`

src/stories/pages/Login/Login.js
```diff
@@ -0,0 +1,134 @@
...
+
+    service.getUsers().then((users) => {
+      try {
+        users.forEach((item) => {
```
cmutuc25, line 39: Nitpicky but `find()` would make more sense to read. Also, shouldn't the user finding be handled in the backend?

cmutuc25 (approved):
I added some comments that are more questions and nitpicks than actual changes.

src/stories/pages/Login/Login.js
alecherryy: I forgot it there lol. Thanks for catching it!

src/stories/pages/Login/Login.js
alecherryy: @cmutuc25 there's a `findUserById` in the BE, however, we don't have access to that when a user first logs onto the platform.

alecherryy: Also, great suggestion! So much cleaner.

alecherryy:
I merged master into this branch and re-tested the Login. All is good on my end.

PR #35, opened 2021-03-01
SIE-62 created user class with getters and setters

dustinlo:
It is intended to only pass in email and password to the constructor at this point.

### DIFF
--- a/src/firebase/CRUDServices/user.js
+++ b/src/firebase/CRUDServices/user.js
@@ -1,0 +1,114 @@
+export default class User {
+    /**
+     * Private fields for user object.
+     */
+    #id;
+    #username;
+    #password;
+    #firstName;
+    #lastName;
+    #email;
+    #role;
+    /**
+     * A constructor to create a User object.
+     *
+     * @param {String} email email of the user.
+     * @param {String} password password of the user.
+     */
+    constructor(email, password) {
+      this.#email = email;
+      this.#password = password;
+    }
+    /**
+     * Setter for id of the user.
+     * @param {int} id user id of the user.
+     */
+    set id(id) {
+      this.#id = id;
+    }
+    /**
+     * Return the id of this user.
+     */
+    get id() {
+      return this.#id;
+    }
+    /**
+     * Setter for username of the user.
+     * @param {String} username username of the user.
+     */
+    set username(username) {
+      this.#username = username;
+    }
+    /**
+     * Return the username of this user.
+     */
+    get username() {
+      return this.#username;
+    }
+    /**
+     * Setter for password of the user.
+     * @param {String} password password of the user.
+     */
+    set password(password) {
+      this.#password = password;
+    }
+    /**
+     * Return the password of this user.
+     */
+    get password() {
+      return this.#password;
+    }
+    /**
+     * Setter for firstName of the user.
+     * @param {String} firstName firstName of the user.
+     */
+    set firstName(firstName) {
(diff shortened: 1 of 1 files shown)


nikrasiya (approved):

src/firebase/CRUDServices/user.js
```diff
@@ -0,0 +1,114 @@
...
+    /**
+     * Private fields for user object.
+     */
+    #id;
+    #username;
+    #password;
+    #firstName;
+    #lastName;
+    #email;
+    #role;
```
quanganhhoang, line 11: What does the #variable mean?

src/firebase/CRUDServices/user.js
```diff
@@ -0,0 +1,114 @@
+export default class User {
```
quanganhhoang, line 1: This kinda java oop/getter/setter code seems a little weird to me. I'd like more people to look at this please @harmanwho @cmutuc25 @alecherryy

quanganhhoang (changes requested):

src/firebase/CRUDServices/user.js
alecherryy: @quanganhhoang it appears `#` is the proposed JS syntax for private fields in a Class. However, this is not standard yet and according to the [docs](https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Classes/Private_class_fields) it's not widely supported.

src/firebase/CRUDServices/user.js
cmutuc25: As far as I know, this isn't a thing in React js. We should probably talk about how we're handling data @harmanwho @alecherryy whenever we meet tomorrow.

dustinlo: Please let me know about it. I can definitely change them to the preferred way.

src/firebase/CRUDServices/user.js
dustinlo: I thought I have to use this style since this was out more than a year ago. Didn't think of that react might not support this. I can revise this to the preferred way.

alecherryy (commented):
@dustinlo let us chat about it tomorrow and we'll let you know!

src/firebase/CRUDServices/user.js
```diff
@@ -0,0 +1,114 @@
...
+    }
+    /**
+     * Return the email of this user.
+     */
+    get email() {
+      return this.#email;
+    }
+    /**
+     * Setter for role of the user.
+     * @param {int} role user role of the user. 0 represents participants.
```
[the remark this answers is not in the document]
dustinlo: Are you suggesting just using string like 'participant' or 'admin'? I just thought by using string literal, typo or upper lower case may be a problem? Thought of using enum but after I searched about it, there is not a good way to do so like Java? (maybe I am just not good at finding things out on JavaScript.) 
@bleachalon What do you think?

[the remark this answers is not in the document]
bleachalon: I think you both have a valid point. If you decide to use an integer, I would suggest changing the name to a more semantic one such as permission level or authorization level instead of role.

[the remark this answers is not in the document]
dustinlo: I will revise them all together after hearing back from your meeting tomorrow.

dustinlo:
I'll close this pull request for now and re-submit later if necessary.

PR #39, opened 2021-03-03
Feature/sie56 modal

harmanwho:


src/stories/components/Modal/Modal.js
```diff
@@ -0,0 +1,84 @@
...
+        isButton={true}/>
+      <div
+        hidden={!modalOpen}
+        id='modalPop'
+        className={['modal', `modal--${theme}`].join(' ').trim()}
+      >
+        <div
+          className={
+            (modalOpen) ?
+              'modal--content modal--open' :
```
alecherryy, line 50: @harmanwho classes for child elements should be `modal__el`; replace this with `modal__content`.

src/stories/components/Modal/Modal.js
```diff
@@ -0,0 +1,84 @@
...
+              'modal--content'}
+        >
+          <span
+            onClick={toggleModal}
+            className='modal--content--close'
+          />
+          {children}
+        </div>
```
alecherryy, line 58: Semantically speaking this should be a button. You can have text inside of it, visually hide it via CSS and use the pseudo-element to include the icon. This is the code snippet I use:
```
display: block;
color: transparent;
font: 0/0 a;
height: 24px;
text-shadow: none;
width: 24px;
position: relative;
```
It visually hides the text, but keeps it accessible to screen readers. It's pretty neat! Let me know if you have any questions!

src/stories/components/Modal/styles.scss
```diff
@@ -0,0 +1,50 @@
...
+/* The Modal (background) */
+.modal {
+  position: fixed;
+  z-index: 1;
+  left: 0;
+  top: 0;
+  width: 100%;
+  height: 100%;
+  overflow: auto;
+  // background-color: rgba(0,0,0,0.4);
```
alecherryy, line 15: You can un-comment this and use this RGB value: `rgba(255, 255, 255, 0.9)`

alecherryy (changes requested):

src/stories/components/Modal/Modal.js
harmanwho: @alecherryy That's a great idea, I did not know that!
I tried it but I wasn't able to hide the button; an empty block of sorts still appears.
Screen Shot 2021-03-03 at 2 23 22 PM

alecherryy (approved):

src/stories/components/Modal/Modal.js
```diff
@@ -0,0 +1,88 @@
...
+        <div
+          className={
+            (modalOpen) ?
+              'modal__content modal--open' :
+              'modal__content'}
+        >
+          {/* <span
+            onClick={toggleModal}
+            className='modal__content--close'
+          /> */}
```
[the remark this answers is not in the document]
harmanwho: No, I didn't. Sorry

PR #41, opened 2021-03-03
Feature/sie 37 create signup page

stenizwahyudi:
I'm not sure about the firebase for signup. I know that the backend is not completed as of now, so I borrowed code from login page for sending the user service for signup and the tests.


src/stories/pages/Signup/Signup.js
```diff
@@ -0,0 +1,131 @@
...
+                type='password'
+                placeholder='Password'
+                handleChange={(e) => setPassword(e.target.value)}
+              />
+              <FormItem
+                modifierClasses={isDarkTheme ? 'form-item--light' : ''}
+                label='Role'
+                type='select'
+                options={['Researcher', 'Participant']}
+              />
```
harmanwho, line 95: I think it will be better to not hard code these values, and declare them using `useState`. Also, have another option - `--Choose--User--` something which isn't valid so that the user doesn't accidentally signup researcher if they don't intend to.

harmanwho (commented):
Everything else looks really good!

alecherryy:
@stenizwahyudi Based on the updated BE, when a user signs up, we want to grab the following information: 
1. First Name
2. Last Name
3. Email
4. Password
5. Role
Right now, you're only passing the email and password to the back-end. You will have to update the form and states.

alecherryy (commented):
Please, make the updates mentioned in my previous content.

venjason (commented):
Overall everything looks great! I think Harman and Alessia both made valid points that we should consider. Something else that I noticed was that the "Dark Theme" isn't technically "dark". It's great that it becomes a darker shade of teal, but the entire left side stays white. 

Wondering if it would be a good idea for "Dark Theme" left hand side to become dark grey / black with white font?

alecherryy:
> Overall everything looks great! I think Harman and Alessia both made valid points that we should consider. Something else that I noticed was that the "Dark Theme" isn't technically "dark". It's great that it becomes a darker shade of teal, but the entire left side stays white.
> 
> Wondering if it would be a good idea for "Dark Theme" left hand side to become dark grey / black with white font?

Oh the theme parameter affects the gradient background and the form colors. We can probably figure out a better name.

src/stories/pages/Signup/Signup.js
stenizwahyudi: I've changed the values of the select dropdown using useState and added 'Select User' as one choice that would be invalid. However, I couldn't get the selected value to update the role for some reasons. I'm pretty sure I'm missing something here. If any of you could help, that would be awesome.

stenizwahyudi:
> @stenizwahyudi Based on the updated BE, when a user signs up, we want to grab the following information:
> 
> 1. First Name
> 2. Last Name
> 3. Email
> 4. Password
> 5. Role
> 
> Right now, you're only passing the email and password to the back-end. You will have to update the form and states.

I have fixed this issue. Last night I checked that user only has email and password, so that's why I only used those 2 information, but I guess I was wrong.

alecherryy (approved):

PR #43, opened 2021-03-04
SIE-66-Add-New-Experiment

harmanwho:
Missing parts:
1. Tests
2. Saving experiment in firebase

### DIFF
--- a/src/stories/pages/AddExperiment/AddExperiment.js
+++ b/src/stories/pages/AddExperiment/AddExperiment.js
@@ -11,17 +11,14 @@ import { Modal } from '../../components/Modal/Modal';
 
 /**
  * Component for AddExperiment page.
- * @param {string} theme Theme of the parent component.
  * @param {string} buttonText Text to appear on the modal's open button.
  * @component
  * @return {object} (
- *   <AddExperiment theme={theme}
- *                  buttonText={buttonText}
- *   />
+ *   <AddExperiment buttonText={buttonText} />
  * )
  */
 
-export const AddExperiment = ({ theme, buttonText }) => {
+export const AddExperiment = ({ buttonText }) => {
   const [error, setError] = useState('');
   const [title, setTitle] = useState('');
   const [shortDesc, setShortDesc] = useState('');
@@ -66,7 +63,6 @@ export const AddExperiment = ({ theme, buttonText }) => {
 
   return (
     <Modal
-      theme={theme}
       buttonText={buttonText}
     >
       <h3>Add New Experiment</h3>
@@ -114,13 +110,13 @@ export const AddExperiment = ({ theme, buttonText }) => {
           handleChange={(e) => setPsychUrl(e.target.value)}
         />
         { error &&
-                <div className="form__msg">
-                  <p>{error}</p>
-                </div>
+          <div className="form__msg">
+            <p>{error}</p>
+          </div>
         }
         <Button
           text="Add Experiment"
-          modifierClasses="button--small button--secondary"
+          modifierClasses="button--tertiary"
           isButton={true}
           onClick={postExperiment}
         />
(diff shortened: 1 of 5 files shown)


src/stories/pages/AddExperiment/AddExperiment.js
```diff
@@ -0,0 +1,137 @@
...
+ * )
+ */
+
+export const AddExperiment = ({ theme, buttonText }) => {
+  const [error, setError] = useState('');
+  const [title, setTitle] = useState('');
+  const [shortDesc, setShortDesc] = useState('');
+  const [longDesc, setLongDesc] = useState('');
+  const [qualtrics, setQualtrics] = useState('');
+  const [spych, setSpych] = useState('');
```
fernandowinfield, line 28: Does this refer to jsPsych? If it does maybe we should change this to `psych` and `setPsych`?

src/stories/pages/AddExperiment/AddExperiment.js
```diff
@@ -0,0 +1,137 @@
...
+import { Button } from '../../components/Button/Button';
+import { Modal } from '../../components/Modal/Modal';
+
+/**
+ * Component for AddExperiment page.
+ * @param {string} theme theme of the parent component.
+ * @param {string} buttonText text to appear on the open button.
+ * @component
+ * @return {object} (
+ *   <AddExperiment isDarkTheme={isDarkTheme}>
```
fernandowinfield, line 18: To keep things consistent I would suggest updating the parameters in this comment to match the params above (`theme` and `buttonText`) instead of `isDarkTheme`

src/stories/pages/AddExperiment/AddExperiment.js
```diff
@@ -0,0 +1,137 @@
...
+          placeholder="SpychJS URL"
+          type="text"
+          showLabel={false}
+          label="SpychJS URL"
+          handleChange={(e) => setSpych(e.target.value)}
+        />
+        { error &&
+                <div className="form__msg">
+                  <p>{error}</p>
+                </div>
```
fernandowinfield, line 110: Maybe in the future, if time allows, we can use an Alert here or add a bit of styling to highlight the error message

src/stories/pages/AddExperiment/AddExperiment.js
```diff
@@ -0,0 +1,137 @@
...
+      shortDesc,
+      longDesc,
+      qualtrics,
+      spych,
+    };
+
+    // To be removed as firebase ID should be automatically generated
+    const experimentId = Math.random();
+
+    service.postExperiment(experimentId, experiment).then((response) => {
```
fernandowinfield, line 46: I guess if the ID is not needed by Firebase we could just remove it from here? But I guess the API would need to be changed too. If that's the case I think this is fine to leave it as it is

fernandowinfield (changes requested):
PR looks great to me in general! Just added some nitpick suggestions on how to rename a few things.

src/stories/pages/AddExperiment/AddExperiment.js
```diff
@@ -0,0 +1,137 @@
...
+      buttonText={buttonText}
+    >
+      <h3>Add New Experiment</h3>
+      <p>Make sure all fields are filled out, submit the new experiment
+        and test it yourself.</p>
+      <Form type="experiment">
+        <FormItem
+          modifierClasses={'form-item--dark'}
+          placeholder="Title"
+          type="email"
```
fernandowinfield, line 70: Forgot to mention: change the type here to "text"?

fernandowinfield (commented):
Forgot to add a comment for an important detail for the FormItem of "Title"

fernandowinfield (approved):

fernandowinfield:
Finished the revision for this PR: SIE-66-Add-New-Experiment.

[Jira Ticket](https://cs6510.atlassian.net/browse/SIE-66)

Changes:
- Renamed variables and slightly changed documentation for consistency and clarity.
- Added prop to button to enable onClick.
- Added a unit test to verify initial rendering.
- Changed `postExperiment()` to handle the right response from the post experiment service.
- Modified the strategy for generating an experiment ID (previously it was a random number, now it is the **title** of the experiment provided by the researcher with any white spaces taken out). For example, if the title of the experiment is "NBA Self Image Experiment", the experimentID would be "NBASelfImageExperiment".

Note:
- Currently, the input fields of the modal do not clear up after an experiment has been added. I am not sure how to do that with the current implementation of the `FormItem` component, but I have a feeling this would be an easy fix if I ask whoever created the FormItem component. If we are expecting researchers to create multiple experiments back-to-back then I think this would be an important thing to do.

alecherryy (approved):

PR #49, opened 2021-03-06
SIE-10-main-menu

venjason:
Creating a pull request for: SIE-10-main-menu 
The menu includes a list of <Link /> components, each link includes:
- An SVG icon
- URL to a page
- Text
Links should include have an is-active class to style the link when the users are on the link’s page. When active:
- Link text is dark teal
- Icon is white
- Teal tab overlaps the icon


src/images/icon-analytics.svg
```diff
@@ -0,0 +1 @@
+<svg xmlns="http://www.w3.org/2000/svg" enable-background="new 0 0 24 24" height="24" viewBox="0 0 24 24" width="24"><rect fill="none" height="24" width="24"/><g><path d="M7.5,21H2V9h5.5V21z M14.75,3h-5.5v18h5.5V3z M22,11h-5.5v10H22V11z"/></g></svg>
```
venjason, line 1: This SVG is not currently in use. Original design has 'analytics', which I believe may be implemented in future so I will leave here.

src/images/icon-reports.svg
```diff
@@ -0,0 +1 @@
+<svg xmlns="http://www.w3.org/2000/svg" height="24" viewBox="0 0 24 24" width="24"><path d="M0 0h24v24H0z" fill="none"/><path d="M16 6l2.29 2.29-4.88 4.88-4-4L2 16.59 3.41 18l6-6 4 4 6.3-6.29L22 12V6z"/></svg>
```
venjason, line 1: Same as above. This SVG is not currently in use.

src/stories/components/MainMenu/MainMenu.js
```diff
@@ -0,0 +1,122 @@
...
+ *   <MainMenu />
+ * )
+ */
+
+export const MainMenu = () => {
+  // List of button links/icons we currently support
+  const icons = ['Dashboard', 'Experiments', 'Account', 'Logout'];
+
+  // Hard coding the color dark teal
+  const darkTeal = '#008489';
```
venjason, line 33: Was unable to figure out how to pull '$teal-dark' from our SCSS file so I've hard coded the color code here.

stenizwahyudi: I think you need to import -> import '../../../scss/variables.scss';
Assign color: '$teal-dark';
Then delete const darkTeal = #008489';

src/stories/components/MainMenu/styles.scss
```diff
@@ -0,0 +1,33 @@
+@import '../../../scss/variables.scss';
+
+// @file
+// Main Menu
+
+// .main-menu {}
+
+.menu-item {
+    min-width: 200px;
+    max-width: 200px;
```
stenizwahyudi, line 10: Not sure if you need to specify the min and max width since I'm pretty sure the component will be used with layout from Sidebar to create the page. Let's wait for other to chime in on this one. It might be okay for now.

src/stories/components/MainMenu/styles.scss
```diff
@@ -0,0 +1,33 @@
...
+	align-items: center;
+    cursor: pointer;
+    font-weight:normal!important; 
+    margin-top: 15px;
+    margin-bottom: 15px;
+}
+
+.menu-item:hover {
+    background: linear-gradient(to right, $teal-dark 35%, white 42%);
+    fill: white;
```
stenizwahyudi, line 22: I think you can separate the icon as separate className, so you can style only the icon using something like  below for hover and active:
        border-radius: 0px 50px 50px 0px;
        background: $teal-dark;

stenizwahyudi: Also, the icons are not lined up properly with the text, the icons appear just a tad higher than the text. Once they are separated into their own class, you can also adjust the position to line up perfectly with the text.

src/stories/components/MainMenu/styles.scss
alecherryy: No need for this!

src/stories/components/MainMenu/MainMenu.js
```diff
@@ -0,0 +1,122 @@
...
+    if (icon === 'Dashboard') {
+      return (<DashboardIcon className='menu-icon' />);
+    } else if (icon === 'Experiments') {
+      return (<ExperimentsIcon className='menu-icon'/>);
+    } else if (icon === 'Account') {
+      return (<AccountIcon className='menu-icon'/>);
+    } else if (icon === 'Logout') {
+      return (<LogoutIcon className='menu-icon'/>);
+    }
+  };
```
alecherryy, line 46: You should probably use a `switch` statement here.

src/stories/components/MainMenu/MainMenu.js
```diff
@@ -0,0 +1,122 @@
...
+
+  return (
+    <Router>
+      <div id='main-menu'>
```
alecherryy, line 50: Why are you using the `id` attribute instead of a class? Also, semantically speaking, you should use a `<ul>` HTML tag and wrap the `<Navlink />` with an `<li>`. This would be the the "correct" structure:
```
<ul className="main-menu">
  <li className="main-menu__item">
    <NavLink className="main-menu__link">Some link</NavLink>
  </li>
</ul>
```

src/stories/components/MainMenu/styles.scss
```diff
@@ -0,0 +1,33 @@
...
+
+}
+.menu-item:active, .active {
```
alecherryy, line 25: You should nest the `.active` class inside the `menu-item`. Otherwise, if another element in the site uses the same class it will inherit the styles.

src/stories/components/MainMenu/styles.scss
```diff
@@ -0,0 +1,33 @@
...
+    background: linear-gradient(to right, $teal 35%, white 42%);
+    fill: white;
```
alecherryy, line 27: Use the global variable `$white`.

src/stories/components/MainMenu/styles.scss
```diff
@@ -0,0 +1,33 @@
...
+    margin-bottom: 15px;
+}
+
+.menu-item:hover {
+    background: linear-gradient(to right, $teal-dark 35%, white 42%);
+    fill: white;
+
+}
+.menu-item:active, .active {
+    background: linear-gradient(to right, $teal 35%, white 42%);
```
alecherryy, line 26: Why are you using a `linear-gradient`? The designs have a shape. I can show you how to style it!

alecherryy (changes requested):
Check your classes, we are using the BEM naming system, so classes should be: `main-menu`, `main-menu__item` and `main-menu__link`.

src/stories/components/MainMenu/MainMenu.js
venjason: Thank you for this template! I forgot to change it back to className. My initial approach was to try and see if I could use the 'id' tag for which is 'active'. But with the help of NavLinks, it simplifies everything so ID's are no longer needed!

src/stories/components/MainMenu/MainMenu.js
venjason: Will do! Thanks :D

src/stories/components/MainMenu/styles.scss
venjason: I will remove it!

src/stories/components/MainMenu/MainMenu.js
venjason: I will try this out! Thanks :D

src/stories/components/MainMenu/styles.scss
venjason: Thank you for advice. Fixed!

src/stories/components/MainMenu/styles.scss
venjason: Fixed!

src/stories/components/MainMenu/styles.scss
stenizwahyudi: I think the .active should be nested inside the menu-item, not the other way around. I saw the updated styles.scss, but I think should be like this:

.menu-item__icon {
    margin-right: 15px;
    padding: 10px 15px 10px 50px;
    display: flex;
    .active {
        background: $teal;
        fill: $white;
        border-top-right-radius: 100px;
        border-bottom-right-radius: 100px;
    }
}

Another thing, maybe we also need to highlight the menu item on hover. I think it'll look better and clearer. As it is not specified in the requirement, maybe you can decide what color goes for the highlighting.

venjason: I totally agree with that idea - 'active' being nested inside of 'menu-item__icon' and 'menu-item__text'. However I was unable / unsuccessful at executing the styles in that fashion. I believe it does not work in that fashion because 'active' is parent to the icon and text. So the only way to utilize the parent 'active' tag is to have the children nested inside of it. 

I have added hover styling to the component!

venjason: 'active' class is unique feature to react router's NavLink. Instead of having a component with state to keep track of which link is active, NavLink is useful in that it will automatically keep track of which 'link' is active, and adds the 'active' class to that element. This 'active' tag can be renamed via activeClassName but it does not change the structure / architecture of the parent to child relationship.

stenizwahyudi: @alecherryy can I get your second opinion on this? I think everything else looks great, just want to make sure about this one. Thanks!

venjason: Yes please! I am quite new to front end styling so I may be wrong... but that's the conclusion I came to after spending hours of trying to figure it out 😅

cmutuc25: I think the current implementation is fine, if you look at how the `active` tag is applied, it's applied to the `parent` component of `NavLink` so I don't think you can nest it

venjason: Update: with the help of Carlo, I have changed the styles so that the active class is nested into the parent element! The active class will not interfere with other components now if an active class is used!

stenizwahyudi: That makes sense! Thank you @cmutuc25 for checking!

cmutuc25 (approved):

PR #51, opened 2021-03-10
SIE-93 Implemented firebase backend API [first draft]

quanganhhoang:
This is my first attempt to implement the firebase backend API. I'm not requesting to change current existing code but only offering this PR as an alternative. I'm fine either way, as long as it makes sense for frontend & backend to work together. 

**What's in this PR?**

1. Refactored existing code
2. Added function to create a user document in `users` collection upon sign-up
3. Ensure proper returns per design doc `{ status, data, error }`. Also added `http-status-code` package to display status codes.
4. Break up a few methods into smaller ones for more flexibility
5. Moved all constants into `constants.js`

**Test**:
- I have not tested any of this code yet. Could anyone please show me how or maybe we could break this into tasks and have each of us test one?

**How to review?**
- It's best to test each and every function but since we don't have time this sprint, let's move it to the next?
- Code owners: please make sure my refactoring is okay and whether you are okay with the changes / apologies for moving code around ~
- Please make sure status codes are correct and my logic is sound
- I most likely made a lot of syntax errors ~~~ Please make sure the javascript code is syntactically correct, especially with async stuff. I'm a big noob here.

NOTE: Most of these methods have already been implemented by @bleachalon, @theodore3131 and @dustinlo. I have only revised and refactored those. 

[Design doc](https://docs.google.com/document/d/1gQcZ3XtKMkwFgy5V2GJut4Nj00HLVZDEBRIMgGCk8_E/edit#heading=h.fas6p7dy44js) written by @dustinlo and @RaccoonQ.


bleachalon:
I don't quite understand why you create another user.js. Do you mind tell us more about the purpose of your functions?

src/firebase/api/users.js
```diff
@@ -0,0 +1,115 @@
...
+export const getCurrentUser = () => {
+  // resolve: callback function if resolved
+  // reject: callback function if rejected
+  return new Promise((resolve, reject) => {
+    const unsubscribe = auth.onAuthStateChanged((userAuth) => {
+      unsubscribe();
+      resolve(userAuth);
+    }, reject);
+  });
+};
```
cmutuc25, line 94: This is interesting, so on Promise resolved, it unsubscribes itself?

src/firebase/api/gcp-utils.js
```diff
@@ -0,0 +1,79 @@
...
+  gcpPubsub.topic(pubsub.SIE_RESULT).onPublish((msg) => {
+    // Decode the PubSub Message body.
+    const msgBody = msg.data ?
+      Buffer.from(message.data, 'base64').toString() : null;
+    // eslint-disable-next-line no-unused-vars
+    const participantId = msg.attributes.participant_id;
+    // eslint-disable-next-line no-unused-vars
+    const experimentId = msg.attributes.experiment_id;
+
+    if (msgBody === 'completed') {
```
quanganhhoang, line 48: RFC: what should we do here?

fernandowinfield: I am not part of the reviewers for this PR but I figured I could provide some ideas.

In my opinion, since this is an "API", I think returning something would be a good solution rather than using a hook. But a disclaimer here is that I have basically 0 experience with React hooks so I could be wrong.

In terms of that to return, I think there could be a few options, for example:

1. Return a success code such as 200 (OK) for when completed and 204 (NO_RESPONSE) for when still not completed.
2. Return some message that looks like: `{message: "completed", fileUrls: [<urls go here>]}` for when completed and `{message: "not_completed", fileUrls: []}` for when still not completed.

Note that for option #2, the `fileUrls` list doesn't really need to be part of the response of the `observeStimuliCompletion()` API. I just put it in there as an idea that could save us an additional call to `getFileUrlsFromBucket()`.

quanganhhoang: yea I thought about that but I'm not sure because this is basically subscribing to a pubsub topic so technically it is not returning anything, more like execute this action if you receive a message kinda thing. In this particular case, we should only let participants start the image selection phase once they have finished all prior steps && a completion message above is received.

theodore3131: Yeah so just a little clarification,  this method is supposed to be an **ongoing active listener**, but not a function to be called when the frontend wants to get processed images.

so once it has got the "completed" message, it will 
- fetch stimuli image URLs from the cloud storage bucket(SIE_STIMULI_IMGS)
- store those image URLs into the firestore collection ("ProcessedImages" collection)

More like an automation process running.

For frontend to get the stimuli images: I used the imageService.js to do CRUD operation for the "ProcessImages" collection and the frontend can call the getProcessedImages  to fetch those URLs from firestore directly.

src/firebase/constants.js
```diff
@@ -0,0 +1,13 @@
+export const firestoreCollections = {
+  EXPERIMENT: 'experiments',
+  USER: 'users',
+};
```
theodore3131, line 4: Add a 
ProcessedImages: 'images' 
probably?

src/firebase/api/gcp-utils.js
quanganhhoang: @cmutuc25 fyi
@theodore3131 why do we need to store the stimuli in a firestore collection?

theodore3131: So if the stimuli images url is only used at the next step of the image selection task, we can definitely just use react hook function setState() to store them for use next step. I was assuming that if we may use it in any future steps, we can easily retrieve them from firestore collection. But if not we can just simply go with react hook function.

quanganhhoang: yea I think we can just cache them in browser

theodore3131: Oh I also realized that since this is a background listening function, we also need to check the participantID and experimentID is align with the current one to get the correction message.  

So I think we still need the frontend call this function passing the participantID and experimentID. I can push the commit to this branch if you want.

quanganhhoang: so I did it a little differently here. I don't want every user to have to check every message, rather there should be a way for us to parse the message and signal the corresponding participant to start the survey. Let's see how the frontend wants to do this.

theodore3131: Sure, so I guess this is also one of the reason why I use the firestore. We can use the firestore to store the data into the corresponding participant without checking the message to align with the current one.

However, this also brings another concern:

if two users doing it at the same time, there will be two subscribers,
they will get the same message, say 'user1-001-complete' 
then via this way, it will store the same URLs **twice** into the firestore as two users call this.

But yeah, let's talk with the frontend team to decide how to deal with this.

src/firebase/api/experiments.js
```diff
@@ -0,0 +1,142 @@
+import firestore from '../firebase.js';
```
cmutuc25, line 1: This import is wrong, it should be `{ firestore }`. You're actually import `firebase` with this and renaming it to `firestore`

src/firebase/api/users.js
```diff
@@ -0,0 +1,145 @@
+import firestore from '../firebase.js';
+import auth from '../firebase.js';
```
cmutuc25, line 2: Same import issue as before, should be `{ firestore, auth }`

src/firebase/api/experiments.js
quanganhhoang: ah duh! Thanks for spotting it!

src/firebase/api/gcp-utils.js
fernandowinfield: Oh oops, I wasn't that was more of an ongoing listener. Thanks for explaining that!

So, how exactly would the React hook alternative work? I am visualizing that once the `observeStimuliCompletion()` gets the 'completed' message, then it could update the state or somehow let the frontend component know that the stimuli images are ready and when the frontend gets that cue then we go ahead and use the `getFileUrlsFromBucket()` API to fetch the images and go about the experiment?

If it's cumbersome to answer my questions it's all good, I can just ask other people in the FE team later to explain this to me 👍

quanganhhoang: yea that's exactly how I'm envisioning it, not so sure about the details yet

harmanwho (approved):

PR #53, opened 2021-03-10
Feature/sie 55 create user greeting

stenizwahyudi:
Here is the link to the board:
https://cs6510.atlassian.net/secure/RapidBoard.jspa?rapidView=4&projectKey=SIE&modal=detail&selectedIssue=SIE-55


src/stories/components/UserGreeting/UserGreeting.js
```diff
@@ -0,0 +1,54 @@
...
+  const service = UserServices.getInstance();
+  const user = useContext(UserContext);
+  const [logInState, setLogInState] = useState({ isLoggedIn: false });
+  const [firstName, setFirstName] = useState('');
+  const [lastName, setLastName] = useState('');
+  const [error, setError] = useState('');
+
+  useEffect(async () => {
+    service.getCurrentUser(setLogInState).then((response) => {
+      if (response.errorCode) {
```
fernandowinfield, line 26: I am not 100% sure, but I think we might have to be extra careful with expecting a response from the `getCurrentUser()` API. I looked at the code from that API and it looks like it doesn't return anything if the call to that API is successful (when the user is logged in and authenticated).

So maybe you could change line 26 to first check if there is a response before checking if there is an errorCode? Something like:

```
if (response && response.errorCode) {
  ...
}
```

I believe that statement will first check if the response object even exists, and only if it does exist then it will check for the errorCode.

By the way, this might be related to the Travis build fail that this PR is getting because I saw this message somewhere in there: `UnhandledPromiseRejectionWarning: TypeError: Cannot read property 'errorCode' of undefined`.

src/stories/components/UserGreeting/UserGreeting.js
```diff
@@ -0,0 +1,54 @@
+import './styles.scss';
+
+import React, { useState, useEffect, useContext } from 'react';
+import UserServices from '../../../firebase/CRUDServices/userServices';
+
+export const UserContext = React.createContext();
```
fernandowinfield, line 6: I wonder if we could get rid of this and just get the context value directly in line 18. I think that should work as long as the name of the context (in this case `UserContext`) is the same as the context created higher in the tree hierarchy. I saw that @cmutuc25 suggested moving the `createContext()` up to the App level component, and I agree with that. Maybe that could be done in this PR or maybe I could put up a quick separate PR to do that. Regardless, I think we could remove this and should be able to get the context value below.

src/stories/components/UserGreeting/UserGreeting.js
```diff
@@ -0,0 +1,54 @@
...
+    <div>
+      { logInState.isLoggedIn ?
+        <UserContext.Provider value={logInState.user}>
+          <div>
+            <h5>Hello, {firstName + ' ' + lastName}</h5>
+          </div>
+        </UserContext.Provider> :
+        <div>
+          { error }
+          <p>No user is logged in</p>
```
fernandowinfield, line 49: In my opinion, the page of the website would look better if we don't show this when no user is logged in. I would say it would be better to display a link to the Login page instead. Or for now, to keep it simple we could just not display anything when no one is logged in. But that's just my guess, maybe we could check with @alecherryy or @cmutuc25?

src/stories/components/UserGreeting/UserGreeting.test.js
```diff
@@ -0,0 +1,46 @@
...
+    // getcurrentUser return not logged in user
+    const mockGetInstance = jest.fn().mockReturnValue({
+      getCurrentUser: (cb) => Promise.resolve(cb({ isLoggedIn: false })),
+    });
+    // set implementation to mock implementation
+    UserServices.getInstance = mockGetInstance;
+    // render component to test
+    render(<UserGreeting />);
+    // assertions to test
+    expect(screen.getByText('No user is logged in')).toBeInTheDocument();
```
fernandowinfield, line 43: If we decide to get rid of the text shown when no user is logged in, then I think this test would need to be slightly modified. Is there such a thing as `toNotBeInTheDocument()`? Because I think could be a solution.

Regardless, I think these tests look great! 😃

fernandowinfield (changes requested):
Left comments for 3 main topics:

- The handling of the response from the `getCurrentUser()` API call.
- Removing the `createContext()` call.
- Suggestion for the text displayed when no one is logged in.

Besides those, this PR looks good to me. The only other thing is that it is currently failing the Travis build tests so you might have to check the build failure details if my suggestion doesn't fix that. Another thing I read in the build failure messages was that there might be a small linter error/warning with some unused variable or something like that...

src/stories/components/UserGreeting/UserGreeting.js
stenizwahyudi: I agree with you. I saw that @cmutuc25 suggested to move the UserContext to the higher tree hierarchy, but I think maybe I still need to include this in this PR for now.

src/stories/components/UserGreeting/UserGreeting.test.js
```diff
@@ -0,0 +1,46 @@
+import { findByText, render, screen } from '@testing-library/react';
```
denglish7, line 1: I think findByText is the unused variable that Fernando mentioned

stenizwahyudi: Good catch! Thank you

src/stories/components/UserGreeting/UserGreeting.test.js
stenizwahyudi: I think there is .not.toBeInTheDocument()

src/stories/components/UserGreeting/UserGreeting.js
stenizwahyudi: I think I'll just show the error message if no user is found for now. When it is integrated with the page, I think it will be handled differently, such as not redirected to the page when login is not authenticated.

fernandowinfield:
Let's just fix the Travis build that's failing and this will be good to go from my point of view. Here are some hints I gathered from the error messages:

```
/home/travis/build/neu-self-image-experiments/sie-firebase/src/stories/components/UserGreeting/UserGreeting.test.js
   8:1  error  Expected indentation of 4 spaces but found 2  indent
  22:1  error  Expected indentation of 6 spaces but found 4  indent
✖ 2 problems (2 errors, 0 warnings)
  2 errors and 0 warnings potentially fixable with the `--fix` option.
```

and

```
   console.error
      Warning: An update to UserGreeting inside a test was not wrapped in act(...).
      
      When testing, code that causes React state updates should be wrapped into act(...):
      
      act(() => {
        /* fire events that update state */
      });
      /* assert on the output */
      
      This ensures that you're testing the behavior the user would see in the browser. Learn more at https://reactjs.org/link/wrap-tests-with-act
          at UserGreeting (/home/travis/build/neu-self-image-experiments/sie-firebase/src/stories/components/UserGreeting/UserGreeting.js:18:16)
      30 |       } else {
      31 |         setError('');
    > 32 |         setLogInState({ isLoggedIn: true });
         |         ^
      33 |         setFirstName(user.firstName);
      34 |         setLastName(user.lastName);
      35 |       }

```

src/stories/components/UserGreeting/UserGreeting.js
stenizwahyudi: Thank you for checking on this. I didn't pay too much attention on what the API is returning when error is thrown.

stenizwahyudi:
Dang it, finally got the build passed by Travis. There is definitely something wrong with me today, LOL.

fernandowinfield (approved):

PR #60, opened 2021-03-18
define callback for getting images & fix upload

theodore3131:
- Define callback function for getting processed images from cloud storage
- Fix the image upload destination issue
    - should be firebase.app().storage() instead of firebase.storage
- Update the error handling

Files to review:
- gcp-util.js
- firebase.js

### DIFF
--- a/src/firebase/api/gcp-utils.js
+++ b/src/firebase/api/gcp-utils.js
@@ -1,32 +1,31 @@
-import { storage, pubsub as gcpPubsub } from '../firebase.js';
-import { pubsub } from '../constants.js';
+import { app, pubsub as gcpPubsub } from '../firebase.js';
+import { pubsub, storageBuckets } from '../constants.js';
 
 /**
  * Upload an image to the specified bucket
  * @param {string} userId userId
  * @param {string} experimentId experimentId
  * @param {Blob} image image file to be uploaded
- * @param {string} bucket destination bucket
  *
  * @return {JSON}
  */
 export const uploadImageToStorage = async (
-  userId, experimentId, image, bucket) => {
+  userId, experimentId, image) => {
   const imagePath = `${userId}-${experimentId}/${image.name}`;
 
-  const storageRef = storage.ref();
-  const rawImageBucketRef = storageRef.child(bucket);
+  const rawImageBucketRef = app.storage(storageBuckets.SIE_RAW_IMGS).ref();
   const newImageRef = rawImageBucketRef.child(imagePath);
-  newImageRef.put(image).then((snapshot) => {
+  return await newImageRef.put(image).then(() => {
     return {
-      status: '201',
-      error: null,
+      status: 201,
+      message: 'image successfully uploaded',
+      data: {},
     };
-  }).catch((error) => {
+  }).catch((err) => {
     return {
-      status: '500',
-      data: null,
-      error,
+      status: 401,
+      message: `not authenticated ${err}`,
+      data: {},
     };
   });
 };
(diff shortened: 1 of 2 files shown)


src/firebase/api/gcp-utils.js
```diff
@@ -1,74 +1,82 @@
...
 /**
  * Get all file urls from bucket to display in img tags
- * @param {string} bucket storage bucket
- * @param {string} bucketPrefix prefix to destination storage directory
+ * @param {String} userId user id
+ * @param {String} experimentId experiment id
+ * @return {String[]} array of processed self image urls
  */
-export const getFileUrlsFromBucket = async (bucket, bucketPrefix) => {
+const getFileUrlsFromBucket = async (userId, experimentId) => {
```
alecherryy, line 67: Is this function missing a return statement? I see URLS being pushed to the array, but the array is not being returned.

src/firebase/api/gcp-utils.js
```diff
@@ -1,74 +1,82 @@
...
+
+  await stimuliImagesRef.getFiles((err, files) => {
     if (!err) {
       files.forEach((file) => {
         file.getDownloadURL().then((url) => {
           fileUrls.push(url);
-        }).catch((error) => {
-          // TODO(qhoang) error handling
+        }).catch(() => {
+          return [];
```
alecherryy, line 79: Since `fileUrls` is already an empty array, there's no need for this. We can just return `fileURLs`.

src/firebase/api/gcp-utils.js
theodore3131: It is in the end of the function which is auto folded I think.

src/firebase/api/gcp-utils.js
theodore3131: this error is actually cached when **one of** the file download URL is not able to get during the for loop, thus the fileURLs array can have some URLs in it already. 

So shall we return the incomplete array or just an empty array because maybe one of the stimuli images is unable to download?

src/firebase/api/gcp-utils.js
```diff
@@ -1,74 +1,82 @@
...
     return {
-      status: '500',
-      data: null,
-      error,
+      status: 401,
+      message: `not authenticated ${err}`,
+      data: {},
     };
   });
 };
```
quanganhhoang, line 31: I'd just make a new function that is specific to sie-experiments which calls this generic `uploadImageToStorage`. If we are hard-coding this, we will eventually have to do this for new experiments with new buckets anyways.

src/firebase/api/gcp-utils.js
```diff
@@ -1,74 +1,82 @@
...
-  const storageRef = storage.ref();
-  const rawImageBucketRef = storageRef.child(bucket);
+  const rawImageBucketRef = app.storage(storageBuckets.SIE_RAW_IMGS).ref();
   const newImageRef = rawImageBucketRef.child(imagePath);
-  newImageRef.put(image).then((snapshot) => {
+  return await newImageRef.put(image).then(() => {
     return {
-      status: '201',
-      error: null,
+      status: 201,
```
quanganhhoang, line 20: do you mind replacing all these ints with StatusCodes.CREATED, etc.?

src/firebase/api/gcp-utils.js
```diff
@@ -1,74 +1,82 @@
...
+      message: 'image successfully uploaded',
+      data: {},
     };
-  }).catch((error) => {
+  }).catch((err) => {
     return {
-      status: '500',
-      data: null,
-      error,
+      status: 401,
```
quanganhhoang, line 26: likewise

src/firebase/api/gcp-utils.js
quanganhhoang: hmm that's interesting. Does this happen often? We should do a retry (up to 3 times probably) to make sure we get all URLs

src/firebase/api/gcp-utils.js
```diff
@@ -1,74 +1,82 @@
...
  * @return {JSON}
  */
 export const uploadImageToStorage = async (
-  userId, experimentId, image, bucket) => {
+  userId, experimentId, image) => {
   const imagePath = `${userId}-${experimentId}/${image.name}`;
 
-  const storageRef = storage.ref();
-  const rawImageBucketRef = storageRef.child(bucket);
+  const rawImageBucketRef = app.storage(storageBuckets.SIE_RAW_IMGS).ref();
```
quanganhhoang, line 16: does `storage.child(storageBuckets.SIE_RAW_IMGS)` work?

quanganhhoang (changes requested):

src/firebase/api/gcp-utils.js
quanganhhoang: ah i highlighted the wrong one. This is meant for the whole `uploadImageToStorage` function.

theodore3131: Make sense to me, we can export another wrapper function to the frontend then

src/firebase/api/gcp-utils.js
theodore3131: Will update

src/firebase/api/gcp-utils.js
theodore3131: No, if we call firebase.storage() it will only refer to the default bucket, that's why we can see a sie-raw-images folder inside the default bucket now in the console instead of in the sie-raw-images bucket. In order to access the customized bucket, we need to use firebase.app().storage() according to the documentation.

quanganhhoang: TIL! Thank you!

quanganhhoang (approved):
LGTM! Thank you!

EDIT: Could you make a Jira task to add retry or another error handling for the urls fetching? The participants must always be able to start the survey I think, or at least redo a step or something to be able to re-generate and re-fetch all stimuli. I'm not sure what the flow should be though...

theodore3131:
> LGTM! Thank you!
> 
> EDIT: Could you make a Jira task to add retry or another error handling for the urls fetching? The participants must always be able to start the survey I think, or at least redo a step or something to be able to re-generate and re-fetch all stimuli. I'm not sure what the flow should be though...

Thank you! Yeah there is already one Jira task related to this for idle connection error which we will do retry. 
But for now, for the two scenarios of URLs fetching:
1. on the cloud functions side, we will directly pass the two error status in the message to the frontend, "face missing" or "failed"
2. on firebase side, if the **getDownloadUrl** catch errors, we will just return the error message (with an incomplete fileUrls array) from the getFileUrlsFromBucket function, so the frontend can always display the message directly and let the user to redo the previous step. 

So what we can do now is for scenario 2, maybe the frontend can do a retry by calling the **observeStimuliCompletion()** function again before display the error message for incomplete URLs array and let the user redo the previous step?

PR #62, opened 2021-03-24
Feature/sie 116 link pages together

denglish7:
- linked pages together with the Main Menu. 
- finished routing so that if the user is not logged in they are taken to the Login page and Dashboard if they are logged in.
- At the moment, user has to refresh browser after logging in in order to take them to the dashboard.


src/stories/pages/Login/Login.js
```diff
@@ -45,7 +45,7 @@ export const Login = ({ isDarkTheme }) => {
         setError(response.errorMessage);
       } else {
         setError('');
-        window.alert(response.email);
+        // window.alert(response.email);
```
cmutuc25, line 48: We can just get rid of this I think

src/contexts/auth-provider.js
```diff
@@ -0,0 +1,28 @@
...
+const AuthProvider = ({ children }) => {
+  const [user, setUser] = useState();
+  useEffect(async () => {
+    try {
+      const auth = await getCurrentUser();
+      if (auth) {
+        const user = await getUser(auth.uid);
+        setUser(user.data);
+      }
+    } catch (err) {
```
venjason, line 17: Should we alert to log in if there's an error?

venjason (approved):
Everything looks good to me! Dashboard switches between different pages very smoothly.

src/stories/components/MainMenu/styles.scss
```diff
@@ -5,6 +5,8 @@
 
 .main-menu {
   @include list-clean;
+
+  margin-top: calc(#{$spacing-10} * 2);
```
alecherryy, line 9: @denglish7 menu should be in a fixed position on desktop. Apply the styles here instead of `.navigation`.

src/stories/pages/Dashboard/Dashboard.js
```diff
@@ -0,0 +1,85 @@
...
+      <Route path="/experiments">
+        <ExperimentContent />
+      </Route>
+    </Switch>
+  );
+};
+
+const content = 'Here\'s an overview of what\'s going on in your ' +
+  'application. You can review active experiments, check reports, ' +
+  'and analyze real-time data.';
```
alecherryy, line 65: This will probably need to change so that the content is pertinent to what is being displayed.

src/stories/pages/Dashboard/styles.scss
```diff
@@ -0,0 +1,14 @@
...
+
+.dashboard {
+  margin-top: $spacing-12;
+  margin-right: $spacing-3;
+}
+
+.navigation {
+  position: fixed;
+  top: $spacing-1;
+}
```
alecherryy, line 14: Pages shouldn't have any CSS. Also, what is the `margin-right` for? Just curious. If you need to add spacing to the top because of the header, you can do so in the `<Sidebar />` component like so:
```
.sidebar {
  .dashboard & {
    // some styles
  }
}
```

alecherryy (changes requested):
Some minor changes :) thanks!

src/stories/pages/Dashboard/styles.scss
denglish7: Thanks! I'll remove it. The text looked close to the edge of the page so I had added the margin-right.

src/stories/pages/Dashboard/Dashboard.js
denglish7: I was just using this as placeholder content. Should I change this or do you just mean that we'll change it once we get the content for each page?

quanganhhoang:
hey Dan! I fixed the login issue. Though I made several more changes and am documenting it here:

- `Login` now redirects to dashboard upon successful authentication
- Wrap routes in `AuthenticatedRoute` and `UnauthenticatedRoute` for the following scenarios:
  - If unauthenticated -> redirect to login then after login redirect back to where user was (captured by useLocation in react-router-dom)
  - If authenticated and user goes to login/signup -> redirect to dashboard
  - [TODO] If authenticated and in an experiment, redirect to current page whenever user tries to access another page
- Fixed a few bugs here and there
- Refactored `signIn` and have `Login` use the new refactored API

**How to review?**
- `App.js` provides authentication context
- `AuthenticatedRoute` and `UnauthenticatedRoute` contains redirect logic ~ the advantage is that we won't have to do manual redirects
- Authentication context implemented in `contextLib.js`, `App.js` and components which use this context

**Issues**
- Since we added context, our `Login` tests are failing. These are outside of my realm so I need some serious help here. Though I tested manually by navigating the web app and can confirm that my changes work :)

NOTE: I admit I made some big changes here to your code so please feel free to adopt it or revert if it doesn't meet your standards.

src/App.js
```diff
@@ -1,19 +1,62 @@
...
+          </UnauthenticatedRoute>
+          <UnauthenticatedRoute exact path="/signup">
+            <Signup isDarkTheme={false} />
+          </UnauthenticatedRoute>
+          <AuthenticatedRoute exact path="/dashboard">
+            <Dashboard />
+          </AuthenticatedRoute>
+          <AuthenticatedRoute exact path="/experiments">
+            <Experiment />
+          </AuthenticatedRoute>
```
alecherryy, line 54: This is not the correct component. Experiments should point to the screen showing _ALL_ experiments.

src/App.js
```diff
@@ -1,19 +1,62 @@
...
+          <AuthenticatedRoute exact path="/account">
+            <AccountInfoPage />
+          </AuthenticatedRoute>
+        </Switch>
```
alecherryy, line 58: Also, placing Experiments and AccountPage outside the Dashboard is causing the layout to break. The header and Main Menu both disappear. I think they should be moved back inside the Dashboard. Or we can create a new page called "ProtectedContent" or something, where Dashboard, Experiments and Account live.

alecherryy (changes requested):
I am getting an error when rendering the `AccountPage`.

src/firebase/api/users.js
```diff
@@ -39,6 +39,31 @@ export const signUp = async (email, password, userData) => {
   }
 };
 
+/**
```
venjason, line 42: This looks very similar to how I implemented 'signIn' on my branch for refactoring existing APIs to use new ones. I will no longer be creating a pull request for my branch since the changes are being made here instead.

src/stories/pages/Login/Login.js
```diff
@@ -28,24 +29,18 @@ export const Login = ({ isDarkTheme }) => {
...
-    service.signIn(user).then((response) => {
+    signIn(email, password).then((response) => {
       if (response.errorCode) {
         // REDIRECT TO DASHBOARD
         // ERROR HANDLING
         setError(response.errorMessage);
       } else {
-        setError('');
-        window.alert(response.email);
+        setIsAuthenticated(true);
```
venjason, line 43: Getting a TypeError on my end "'setIsAuthenticated' of 'Object(...)(...)' as it is null." Which causes login page to not render.

venjason: My understanding is that 'setIsAuthenticated(true)' means that the user successfully logged in. If thats the case, do we need to still reroute to dashboard after it has set to true?

Also need to move line 39 'REDIRECT TO DASHBOARD' down to the else segment.

cmutuc25: This is outdated now, resolving

alecherryy:
@denglish7 and @cmutuc25 I cannot navigate from the Login page to the Signup page. Also, if I try to go directly to the Signup page, nothing happens. To replicate, go the Homepage/Login page on a private browser window and try to navigate to the Sign up page via the link in the footer.

alecherryy (changes requested):
@denglish7 and @cmutuc25 the navigation between the Login page and the Signup is broken. You can view this, by opening the website in a private browser window.

fernandowinfield:
Copying this from my comment from Slack just for the record:

Login page: clicking on the button seems to log the user in successfully but it doesn’t reroute to `/` or `/dashboard` or whatever page we want to be the landing page after login.

fernandowinfield:
Just ran the branch with the most recent changes. Redirecting after logging in works great now. Same for the links between Login and Signup. The only thing is Experiment and Account (from the left side menu) show a blank page when clicked. Do we only need to add a <Route> for each of those in App.js?

src/contexts/auth-provider.js
cmutuc25: Yeah, we could like show that annoying message

"Username or password is incorrect"

cmutuc25: But that logic could probably go to the Login page, unless we encapsulate that functionality here and return an error with the provider. I think for now we leave it and create a ticket :)

PR #64, opened 2021-03-24
feature/SIE94&95-created-emailverification-and-anonymousUser-plus-minor-fixes

bleachalon:
In this PR:
1. Moved signIn from userService to api/user.js. Added corresponding error handling to incorrect user credentials.
2. Added email verification when signUp and check authentication.
3. Added functionality to create the anonymous user and corresponding user doc.
4. Added functionality to link users who created an account to their anonymous user data.
5. Added signOut user functionality.
6. More detail of the above changes will be added to firebase Doc.

### DIFF
--- a/src/firebase/api/users.js
+++ b/src/firebase/api/users.js
@@ -213,10 +213,10 @@ export const updateUserData = async (updatedData) => {
 };
 
 /**
- * Sign current logged in user out.
+ * Log current out.
  * @return {Object} response object.
  */
-export const signOut = async () => {
+export const logOut = async () => {
   try {
     await auth.signOut();
     return {
--- a/src/firebase/api/users.js
+++ b/src/firebase/api/users.js
@@ -142,7 +142,7 @@ export const getCurrentUser = () => {
   return new Promise((resolve, reject) => {
     const unsubscribe = auth.onAuthStateChanged((userAuth) => {
       unsubscribe();
-      if (userAuth && !userAuth.isAnonymous && userAuth.emailVerified) {
+      if (userAuth && (userAuth.isAnonymous || userAuth.emailVerified)) {
         resolve(userAuth);
       }
     }, reject);
@@ -219,7 +219,7 @@ export const signOut = async () => {
     await auth.signOut();
     return {
       status: StatusCodes.OK,
-      data: userAuth,
+      data: null,
       error: null,
     };
   } catch (error) {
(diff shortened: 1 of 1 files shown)


src/firebase/api/users.js
```diff
@@ -173,3 +177,113 @@ export const updateUserData = async (updatedData) => {
...
+/**
+ * Sign current logged in user out.
+ * @return {Object} response object.
+ */
+export const signOut = async () => {
+  try {
+    await auth.signOut();
+    return {
+      status: StatusCodes.OK,
+      data: userAuth,
```
quanganhhoang, line 222: `data` should be null here? There is no `userAuth`

src/firebase/api/users.js
```diff
@@ -173,3 +177,113 @@ export const updateUserData = async (updatedData) => {
...
+    return {
+      status: StatusCodes.INTERNAL_SERVER_ERROR,
+      data: null,
+      error,
+    };
+  }
+};
+
+/**
+ * This function is for users who do not what to sign in.
```
quanganhhoang, line 235: typo? Can you add more description of when and how to call this function? This feature is super cool though!

src/firebase/api/users.js
```diff
@@ -173,3 +177,113 @@ export const updateUserData = async (updatedData) => {
...
+  }
+};
+
+/**
+ * Link user to their current anonymousAccount.
+ * @param {String} email email of user.
+ * @param {String} password password of user.
+ * @return {Object} userAuth or error object.
+ */
+export const linkUsetToAnonymousAccount = async (email, password) => {
```
quanganhhoang, line 264: typo

quanganhhoang (changes requested):
a few comments but LGTM! Are these tested btw?

src/firebase/api/users.js
bleachalon: > a few comments but LGTM! Are these tested btw?

Yes, they are all tested on main scenarios, but I may miss some corner cases, which I will continue testing. I will fix the typo and errors afternoon.

src/firebase/api/users.js
```diff
@@ -173,3 +177,113 @@ export const updateUserData = async (updatedData) => {
...
+  try {
+    const credential = auth.EmailAuthProvider.credential(email, password);
+    const userAuth = await auth.currentUser.linkWithCredential(credential);
+    return {
+      status: StatusCodes.OK,
+      data: userAuth,
+      error: 'Linking success',
```
theodore3131, line 271: Shall we use 'message' or 'error' for this message? If so we can unify the return format. {status, data, error}

bleachalon: let's add this as a ticket for the next sprint to unify all responses.

src/firebase/api/users.js
bleachalon: Changed it to be more meaningful and will add an example of using this function to the doc.

theodore3131 (approved):

bleachalon:
> a few comments but LGTM! Are these tested btw?

PR #67, opened 2021-03-25
[SIE-122] integrate Participant user journey

alecherryy:
I know the linter is broken. I'm waiting for the last of the work to be merged so I can update the branch, but in the meanwhile, I could use a review. A bunch of this work is simple refactoring. To test the Experiment flow go to [localhost:3000/study/Pl3WJYa7vQ1ALVt0rHRV](localhost:3000/study/Pl3WJYa7vQ1ALVt0rHRV), fill in the form and you should be redirected to the Experiment page where data from the DB for experiment `Pl3WJYa7vQ1ALVt0rHRV` is being pulled in.


fernandowinfield (commented):
I went through the experiment flow that you mentioned (which looks awesome by the way) but didn't know exactly what features to expect to work. That said, here's a few things I saw:

- The 'Next' buttons are not blocked so I was able to go through the wizard steps without filling out any surveys or taking/uploading a photo.
- After clicking 'Next' in step 3 (Photo Instructions and Upload), the wizard moved to step 4 (Pre-Survey) but the window seemed to be scrolled down from the previous step, I couldn't see anything on the screen unless I scroll up. I'm guessing this is an easy fix by programmatically doing some scrolling up?
- No Pre-Study or Post-Study questionnaires showed up (in step 4 and 6) but I think this might be on purpose for now.
- Also probably on purpose, I couldn't find a way to exit the experiment flow.

Besides those and the JsPsych part, everything worked well for me.

src/stories/pages/Experiment/Experiment.js
```diff
@@ -29,8 +29,20 @@ import { ImageSelectionTask }
  */
 
 export const Experiment = ({
-  title, description, consentForms, url, preSurveys, postSurveys,
+  preSurveys, postSurveys,
```
cmutuc25, line 32: I know this isn't implemented yet, but I would think we would be grabbing the surveys through a firebase request in the useEffect below?

cmutuc25:
Closing as this is included in https://github.com/neu-self-image-experiments/sie-firebase/pull/68

PR #72, opened 2021-04-01
[DO_NOT_MERGE] POC for adding jsPsych library

fernandowinfield:
I attempted to use [jspsych-react](https://github.com/makebrainwaves/jspsych-react#readme) but I wasn't able to make it work. So I decided to try to get the basic tutorial of jsPsych working using the library itself directly.

This seems to be working fine but my biggest question is: where would we place the `jspsych-6.3.0` folder (that contains the whole library JavaScript scripts and CSS)? I temporarily placed it in the `public` folder of our React app but I have a feeling there is a much cleaner way to do this. Any thoughts?

Things I did:
- Download the folder (`jspsych-6.3.0`) that contains the whole jsPsych library and put it in `/public`.
- Add the core jsPsych script and their stylesheet in `index.html`.
- Add a placeholder component, `JsPsych`, that loads the script of the plugin used for the experiment and then runs the actual experiment.
- Add a placeholder route to be able to test the experiment. Go to `http://localhost:3000/jspsych` to test it.


fernandowinfield:
I made a mistake and I was loading the plugin script in `index.html`. With this fix I think we actually load the plugin script programmatically when the React component (`<JsPsych/>`) is loaded.

Also, I removed a bunch of unnecessary files from the jsPsych library. So I guess we can only include the plugins that we actually need (plus the core files of the library).

venjason:
> > I attempted to use [jspsych-react](https://github.com/makebrainwaves/jspsych-react#readme) but I wasn't able to make it work. So I decided to try to get the basic tutorial of jsPsych working using the library itself directly.
> > This seems to be working fine but my biggest question is: where would we place the `jspsych-6.3.0` folder (that contains the whole library JavaScript scripts and CSS)? I temporarily placed it in the `public` folder of our React app but I have a feeling there is a much cleaner way to do this. Any thoughts?
> 
> I am gonna take a look at this.

Please let me know if you are successful at using jspsych-react! I will try to see if I can figure it out on my end as well :). 

Not entirely sure, but I think it would make our lives easier if we are able to import the jspsych library rather than using the library itself directly.

cmutuc25:
I think using the whole library would be better, the react package hasn't been updated in years

venjason:
> I think using the whole library would be better, the react package hasn't been updated in years
I think you're absolutely right. I was looking at it earlier and there aren't many resources for it either.

fernandowinfield:
Just replaced the basic tutorial example with the self-image experiment implementation from the previous team. Not sure if 100%, but looks like it works. If you want to check it out, pull this `explore-jsPsych` branch, `npm start`, and go to http://localhost:3000/jspsych.

denglish7:
> Just replaced the basic tutorial example with the self-image experiment implementation from the previous team. Not sure if 100%, but looks like it works. If you want to check it out, pull this `explore-jsPsych` branch, `npm start`, and go to http://localhost:3000/jspsych

Just tried this out, nice work man! It's working for me as well.

PR #73, opened 2021-04-01
[SIE-141] append vars to Consent form

alecherryy:


src/App.js
```diff
@@ -20,6 +20,12 @@ function App() {
   return (
     <BrowserRouter>
       <Switch>
+        <Route path="/study/:experimentId/user/:participantId">
+          <AnonymousLogin />
+        </Route>
+        <Route path="/:experimentId/:participantId">
+          <Experiment />
+        </Route>
```
fernandowinfield, line 28: So, we will have participants land in `<AnonymousLogin />` and then redirect them to <Experiment /> for the actual experiment. Did I understand correctly?

If yes, I am wondering if we should move this Route towards the bottom of the Switch. Maybe I am overthinking here, but for some reason I feel a little uneasy about having a path consist of only URL parameters. For example, if we had a path like `/experiments/details`, would the router mistakenly load `<Experiment />` and think that `experimentId` = 'experiments' and `participantId` = 'details'?

fernandowinfield (approved):
The changes to the code logic make sense and look good to me! I only left a comment about the Experiment route in case it could be important.

The styling changes, I'll be honest, I only breezed through them but I fully trust you that they are great and for a good reason.

src/App.js
alecherryy: Take this as a placeholder. I don't know what kind of format the first URL will have, so I am having a hard time coming up with a good path. I will add a comment noting that it will need to be looked at again as we get closer to finishing the Qualtrics functionality

PR #76, opened 2021-04-01
[SIE-156] Add unit test for gcp-utils

theodore3131:
- Define the test structure for firebase apis
- Write unit tests for gcp-utils.js
- Remove deprecated subscriber.js file


src/firebase/test/gcp-util.test.js
```diff
@@ -0,0 +1,26 @@
...
+  const experimentId = '001';
+  const [urls, setUrls] = useState([]);
+  const [error, setError] = useState('');
+  const imageUrlsHandler = (urls) => setUrls(urls);
+  const errorHandler = (error) => setError(error);
+  await observeStimuliCompletion(userId, experimentId,
+    imageUrlsHandler, errorHandler);
+  expect(urls).toHaveLength(400);
+  expect(error).toBe('');
+});
```
quanganhhoang, line 26: do these tests actually upload an image and generate stimuli?

theodore3131: Not yet, while I am thinking since this unit test will be called every time in Travis CI for any PR, shall we limit this?
Also I noticed that our cloud storage has already exceeded the free 5GB limit.

quanganhhoang: my point is this should not upload an image and generate stimuli since this will no longer be unit test?

> Also I noticed that our cloud storage has already exceeded the free 5GB limit.
ah nice catch. That's filled up so fast. Imma delete some stuff

dustinlo: > my point is this should not upload an image and generate stimuli since this will no longer be unit test?
> 
> > Also I noticed that our cloud storage has already exceeded the free 5GB limit.
> 
> ah nice catch. That's filled up so fast. Imma delete some stuff

We found this in our meeting when we were talking about budget analysis. This seems like the only major cost besides the cost of hosting the site. It will skyrocket the cost if we save the generated stimuli in Firestore.

theodore3131: Yes since every time we upload an image the stimuli is generated, 
I think I can just do a unit test for the function of getting image urls from cloud storage.

dustinlo (approved):

PR #77, opened 2021-04-01
SIE127/128 fetch consent form result and minor fix of get current user

bleachalon:
**Summary:**
1. Created function to fetch stored Qualtrics data from Firestore.
2. Fixed the logic of getCurrentUser which caused infinite loading.

**Discussion**
There is one design uncertainty. Should I create one function to return all user-experiment data, or should I create different functions for each form (consent, post-survey, pre-survey)?


src/firebase/api/qualtrics.js
```diff
@@ -0,0 +1,30 @@
...
+ */
+export const getConsentResult = async (uid, experimentId) => {
+  try {
+    const collectionRef = firestore.collection(`${collections.USER}`)
+      .doc(uid).collection(`${collections.EXPERIMENT}`).doc(experimentId);
+    const snapshot = await collectionRef.get();
+    return {
+      status: snapshot.exists ? StatusCodes.OK : StatusCodes.NOT_FOUND,
+      message: snapshot.exists ? 'sucesss' : 'not found',
+      data: snapshot.exists ? snapshot.data() : null,
```
quanganhhoang, line 21: 1. Let's have a `const snapshotExists = snapshot.exists` so we don't have to call snapshot.exists multiple times. 
2. Typo at `sucesss` and let's use constants here

src/firebase/api/users.js
```diff
@@ -176,7 +176,11 @@ export const getCurrentUser = () => {
   return new Promise((resolve, reject) => {
     const unsubscribe = auth.onAuthStateChanged((userAuth) => {
       unsubscribe();
-      resolve(userAuth);
+      if (userAuth && (userAuth.emailVerified || userAuth.isAnonymous)) {
+        resolve(userAuth);
+      } else {
+        resolve(null);
+      }
```
quanganhhoang, line 183: could you add a comment here explaining what this does?

quanganhhoang (changes requested):

fernandowinfield:
@bleachalon, sorry I was not aware of this PR. The frontend team preference is to have a separate function for each (consent, post-survey, pre-survey). Thanks!

dustinlo (approved):

PR #78, opened 2021-04-02
[SIE-135] Add jsPsych library to project

fernandowinfield:
Small PR to add the library, [jsPsych](https://www.jspsych.org/), that we will use to conduct the trials of the experiment. No functional changes contained in this PR. I put the files in `sie-firebase/src/external_libraries/` but if something like `sie-firebase/src/jspsych/` is a better home for the library I can move it no problem.

Note: in future PRs where we actually add the jsPsych experiment functionality (trials and timeline), we will have to add the script(s) of the jsPsych plugin(s) needed. There are at least two ways to do this:

1. Programmatically insert the scripts right before running the experiment ([see example here](https://github.com/neu-self-image-experiments/sie-firebase/pull/72/commits/9895b5d4b412ff0c9f853664d17052b7b5275033#diff-7e2f6b31103a33591df109ac5c9d69bbb87c6ce5e0fc9d31b2a90fb507cfe945R15-R34)).
2. Directly adding the scripts in `index.html`.

Jira [ticket](https://cs6510.atlassian.net/browse/SIE-135)

### DIFF
--- a/public/index.html
+++ b/public/index.html
@@ -42,6 +42,9 @@
       Learn how to configure a non-root public URL by running `npm run build`.
     -->
     <title>SIE | Scientific Study Platform</title>
+    <!-- jsPsych library core script and styling -->
+    <script src="jspsych-6.3.0/jspsych.js"></script>
+    <link href="jspsych-6.3.0/css/jspsych.css" rel="stylesheet" type="text/css">
   </head>
   <body>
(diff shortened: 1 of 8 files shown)


venjason (approved):
Looks good to me! Per our conversation via Slack - we've proven that adding the jsPysch library is compatible.

PR #79, opened 2021-04-03
[SIE-167] Add api method for storing experiment result

theodore3131:
- Added API method in experiment.js for storing experiment result( **jsPsych data** )
- Unified the import statement

### DIFF
--- a/src/firebase/api/experiments.js
+++ b/src/firebase/api/experiments.js
@@ -1,4 +1,4 @@
-import { firestore } from '../firebase.js';
+import { firestore } from '../firebase';
 import { StatusCodes } from 'http-status-codes';
 
 import {
@@ -129,6 +129,43 @@ export const deleteExperiment = async (experimentId) => {
     // NOTE: this does NOT delete this doc' subcollections
     return {
       status: StatusCodes.OK,
+      data: null,
+      error: null,
+    };
+  } catch (error) {
+    return {
+      status: StatusCodes.INTERNAL_SERVER_ERROR,
+      data: null,
+      error,
+    };
+  }
+};
+
+/**
+ * Store experiment data into specific experiment document of the user.
+ * @param {String} userId user id
+ * @param {String} experimentId experiment id
+ * @param {JSON} resultData json object such as {'jsPsych': data}
+ * @return {JSON} response
+ */
+export const storeExperimentResult = async (userId, experimentId,
+  resultData) => {
+  try {
+    // locate the specific experiment id of the user
+    const userRef = firestore
+      .collection(collections.USER)
+      .doc(userId);
+
+    const expRef = userRef.collection(collections.EXPERIMENT)
+      .doc(experimentId);
+
+    const snapshot = await expRef.get();
+    if (snapshot.exists) {
+      expRef.update(...resultData);
+    }
+
+    return {
+      status: snapshot.exists ? StatusCodes.OK : StatusCodes.NOT_FOUND,
       data: null,
       error: null,
     };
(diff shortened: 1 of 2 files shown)


src/firebase/api/experiments.js
```diff
@@ -140,3 +140,40 @@ export const deleteExperiment = async (experimentId) => {
...
+    const userRef = firestore
+      .collection(collections.USER)
+      .doc(userId);
+
+    const expRef = userRef.collection(collections.EXPERIMENT)
+      .doc(experimentId);
+
+    const snapshot = await expRef.get();
+    if (snapshot.exists) {
+      expRef.update(...resultData);
```
fernandowinfield, line 164: This line overwrites whatever was saved in that experiment collection for that participant, correct?

fernandowinfield (approved):
Clean and simple 👌 .  Just a question: I am just curious if we expect to have any data previously saved in `participantId/experimentId/` in Firebase? Because if we do have some data I think the line I highlighted would replace anything in there with `resultData`?

src/firebase/api/experiments.js
theodore3131: Thanks! No it will only update or add the field of the document instead of overwrite everything according to the [document](https://firebase.google.com/docs/firestore/manage-data/add-data#web_5) Here we only pass a single key:value pair and it shall be added to the user document

venjason: Can I get some clarification on where the experiment result JSON is expected to be stored? I see that we have two collections, 'Experiments' - which stores content data for each individual experiment & 'Users' - which stores user data. My interpretation is that we're expecting to store experiment result data (JSON) inside the 'Users' collection? 

Currently implementing JyPysch trials / experiments and am attempting to store experiment data in Firestore. When I call the API 'storeExperimentResult', it is giving me a 404 error. I am passing in all the variables as expected.

I can kind of understand the logic in this function however am new to firebase. The reason I think I am getting a 404 error is because in line 162, snapshot will never exist since there isn't a field 'Experiments'.

theodore3131: So per our design there shall be a collection of experiments inside the user document.  Thus we can use the experimentID to locate the specific experiment the user is taking, and then store the data in to the experiment document.
So I think while testing, the user should have participate in an experiment first, so that it will have the 
'Experiments' inside to locate the specific expID.

theodore3131: Oh, I think I need to create the 'Experiment' collection inside user Doc with the given experimentID and link it to the existing Experiment when a participant finishes the experiment. I Will update soon, thanks for the reminder!

venjason: Thanks for the prompt response & clarification Theo! 

That makes a lot of sense of where experimental data / future experimental data would be stored. I have altered the logic of storing data slightly. 

`    const snapshot = await expRef.get();
    if (snapshot.exists) {
      // expRef.update(...resultData);
      expRef.set(resultData);
    }
`

Please let me know if this will cause any issues! I believe using set will produce the same expected result & will allow users to replace data / add new row of data for different experiments.

theodore3131: Thanks for the update! Yeah I believe this should be the correct way and I have just opened a PR for this as well, feel free to review and see if it works!

venjason: From testing it locally, I can verify that set works. If you look at ID: Bo0Hka0QT5O2Q026qKpFbAYHOW23 in firebase, there is a 'Experiments' collection within user & contains dummy experimental data produced by JsPsych. I was unable to get update to work.

venjason: https://console.firebase.google.com/u/0/project/cs6510-spr2021/firestore/data~2FUsers~2FBo0Hka0QT5O2Q026qKpFbAYHOW23~2FExperiments~2F9iG2wajKXkjLXiLQfwgj

theodore3131: Cool, thank you for verifying! If it works well we can merge the new PR then.

PR #83, opened 2021-04-06
fix store experiment results bug

theodore3131:
In the PR:
- Fix the Not_found error while the experiment is not defined under the user document


venjason (approved):
Thanks for fixing this. 

More information can found here: https://github.com/neu-self-image-experiments/sie-firebase/pull/79 

Verified that new implementation stores experimental data properly.

PR #85, opened 2021-04-07
[SIE-152] Login page shows feedback

cmutuc25:
Made the login page respond based on the error message received. Error message will show on the page, if something fails, there will be 'Server not responding' message.

Added simple test to check if there's an error message if the backend function returns an error.

Removed some es lint lines.

Refactored auth-provider to provide a reload functionality to make it more obvious how to reload when a user is logged in or changed


fernandowinfield (approved):
Looks good! So now we can just use `useAuth()` instead of `useContext(AuthContext)`?

cmutuc25:
Yes, just so we don't have to import `useContext`. Minor change

cmutuc25:
Need to add a message if a user is has an unverified email

cmutuc25:
Reopening this as i added the email verification error

cmutuc25:
Reopening this as I added the email verification error

PR #86, opened 2021-04-07
Added Logout

harmanwho:
1. AuthProvider: added `resetUserData` function to clear userdata upon logout
2. Added Logout files
3. App.js: Added logout to routes


src/App.js
```diff
@@ -55,12 +55,24 @@ function App() {
...
+          )} />
         <Route exact path="/login">
           {user ? <Redirect to="/dashboard" /> : <Login />}
         </Route>
         <Route exact path="/signup">
           {user ? <Redirect to="/dashboard" /> : <Signup isDarkTheme={true} />}
         </Route>
+        <Route exact path="/logout">
+          {user ? <Redirect to="/logout" /> : <Redirect to="/login" />}
+        </Route>
```
fernandowinfield, line 75: I think we can remove these 3 lines. By using `PrivateRoute` in lines 58-66, that component takes care of the redirecting when a user is not logged in.

src/contexts/auth-provider.js
```diff
@@ -11,6 +11,13 @@ const AuthProvider = ({ children }) => {
   const [loaded, setLoaded] = useState(false);
   const [trigger, setTrigger] = useState(false);
 
+  const resetUserData = () => {
+    setTrigger(false);
+    setUser();
+    setIsAuthenticated(false);
+    setLoaded(false);
```
fernandowinfield, line 18: I think removing this line solves the forever loading problem right after logging out. I believe the reason is because the useEffect that is listening to `trigger` is the one in charge of setting loaded to `true`. If we `setLoaded(false)` after the trigger, then loaded will remain as `false` (and the Loader will stay forever) until the user refreshes the page.

src/stories/pages/Logout/Logout.js
```diff
@@ -0,0 +1,67 @@
...
+
+  const cancelLogout = (e) => {
+    history.push('/dashboard');
+  };
+  return (
+    <Constrain>
+      <HorizontalTitle
+        modifierClasses="horizontal-title--medium"
+        eyebrow="Logout"
+        title="You sure?"
```
fernandowinfield, line 46: Design subjective preference: Change this to `"Logout"`?

src/stories/pages/Logout/Logout.js
```diff
@@ -0,0 +1,67 @@
...
+  };
+
+  const cancelLogout = (e) => {
+    history.push('/dashboard');
+  };
+  return (
+    <Constrain>
+      <HorizontalTitle
+        modifierClasses="horizontal-title--medium"
+        eyebrow="Logout"
```
fernandowinfield, line 45: Design subjective preference: Change this to `"Confirm"`?

src/stories/pages/Logout/Logout.js
```diff
@@ -0,0 +1,67 @@
...
+        title="You sure?"
+        content={<>
+          <Button
+            modifierClasses={'button--small button--secondary'}
+            disabled={true}
+            isButton={true}
+            text="Yes"
+            onClick={(e) => logUserOut(e)}
+          />
```
fernandowinfield, line 54: Design subjective preference: Change the content of the HorizontalTitle to the string `"Are you sure you want to log out?"` and move the 'Yes' and 'No' buttons outside at the same level as the HorizontalTitle (but inside the Constrain)?

fernandowinfield (changes requested):
Two main things. The first one I think we should change because it is a functionality issue. The second thing is subjective and more of a personal preference of the design of the Logout component:

1. Sometime the web app stays on an infinite loading state. To reproduce you can log in, go to the 'Account' tab, then refresh the page, and then log out. See my comments for a suggestion on how to fix this.

2. See my comments with suggestions on how to slightly reorganize/reword the Logout component you just created.

P.s: @cmutuc25 was the mastermind behind the auth-context so maybe he can confirm if my suggested change in auth-provider.js makes sense.

src/contexts/auth-provider.js
```diff
@@ -11,6 +11,13 @@ const AuthProvider = ({ children }) => {
   const [loaded, setLoaded] = useState(false);
   const [trigger, setTrigger] = useState(false);
 
+  const resetUserData = () => {
+    setTrigger(false);
```
fernandowinfield, line 15: I think `setTrigger(!trigger)` solves the forever loading problem right after logging out.

harmanwho: I tried that, it didn't log out but kept redirecting me to the dashboard. It doesn't seem to empty the values.

harmanwho: This bug was in the file `auth-provider`, I removed `setLoaded()` from reset values and it got resolved.

src/App.js
harmanwho: This bug was in the file `auth-provider`, I removed `setLoaded()` from reset values and it got resolved.

src/contexts/auth-provider.js
fernandowinfield: Interesting. It was working for me... We can take a look again tomorrow

fernandowinfield (approved):

src/contexts/auth-provider.js
fernandowinfield: @harmanwho, sorry. I had forgotten to check this again but it looks like it's working well with the last change you made 👍

cmutuc25:
This works, think this is good to go. It's been sitting here awhile as approved so I'm merging

PR #87, opened 2021-04-07
Sie 136 implement js pysch trials

venjason:
Implementing JsPsych using the external library that was previously integrated. 

The experiment functionality works properly. Here are some key features: 
- Experiment successfully grabs the stimuli image URLs from our database via the provided experimentId & participantId.
- Utilizes the back-end api provided and connects with FireBase to update user information to contain experiment Id & experiment data. 
- Responses are accurately recorded when user clicks either 'e' or 'I'.

There are some things to finish where I need some help / input. I will detail more below to pinpoint exactly where the issues are. Here is what I am having difficulties with:
- Rendering the example images (example1.png / example2.png) in the instructions trial.  
- There is a behavior sometimes during the experiment where one image will load/render before the second image. I believe this is the same behavior we saw in the previous class's implementation. 
- Need to update ImageSelectionTask to render JsPsych experiment when users are ready for the experiment. 

To run the experiment locally:
- 'npm start' in console
- Open up a browser and access http://localhost:3000/9iG2wajKXkjLXiLQfwgj/Bo0Hka0QT5O2Q026qKpFbAYHOW23/
- Go to the fourth section where ImageSelectionTask is rendered

[SIE-136 Investigate and implement jsPsych trial type(s)](https://cs6510.atlassian.net/secure/RapidBoard.jspa?rapidView=4&projectKey=SIE&modal=detail&selectedIssue=SIE-136)


src/stories/components/JsPsych/JsPsych.js
```diff
@@ -0,0 +1,240 @@
...
+  // running the experiment; otherwise it crashes.
+  setTimeout(() => {
+    // Implementation of previous team timeline & trial logic -------------------------- START
+    /* create timeline */
+    const timeline = [];
+
+    /* number of trials */
+    // TODO: Adjust NUMBER_OF_TRIALS back to 199
+    // NOTE: Adjust line below to shorten the number of trials. 199 will go through all 200 iterations.
+    const NUMBER_OF_TRIALS = 5;
```
venjason, line 76: Note: we will have to adjust this number back to 199 before launching / going live. I've shortened it to 5 (or a total of 6 trials) for ease of testing at the moment.

src/stories/components/JsPsych/JsPsych.js
```diff
@@ -0,0 +1,240 @@
...
+            "on the screen. Choose the image that looks more like you. </p>" +
+            "<p>Press the letter <strong>E</strong> on the keyboard to select the image on the left.</p>" +
+            "<p>Press the letter <strong>I</strong> on the keyboard to select the image on the right.</p> " +
+            "<p></p>" +
+            "<div style='width: 900px; margin: auto;'>" +
+            // TODO: Figure out how to properly render example images. Both implementations below do not render local images.
+            //        We are able to render via HTTPs URLs but having issues with loading images locally.
+            "<div style='float: left;'><img width='300' src={require('../../../images/example1.png')} alt='Error loading example 1'></img>" +
+            "<p class='small'><strong>Press the E key</strong></p></div>" +
+            "<div class='float: right;'><img width='300' src={ example2 } alt='Error loading example 2'></img>" +
```
venjason, line 93: This is exactly where I've been scratching my head all day. Line 91 - 93. I have been unsuccessful at finding a way to solve this issue. Rendering an image locally. Line 91 and Line 93 are using two different methods of rendering images, both of which do not work.

src/stories/components/JsPsych/JsPsych.js
```diff
@@ -0,0 +1,240 @@
...
+    };
+    timeline.push(instructions);
+
+    /* generate trials with number of trials */
+    function generateTrials(numberOfTrial) {
+      const trials = [];
+      for (let i = 0; i <= numberOfTrial; i++) {
+        const invFilePath = stimuliUrls[i];
+        const oriFilePath = stimuliUrls[i+1];
+        // TODO: Figure out how to render two images at the same time.
```
venjason, line 108: As mentioned in PR, there is a behavior that is present quite consistently, where invFilePath image loads before oriFilePath image. I believe this is occurring because the page is being read top down, meaning invFilePath has more time to render / is able to render first. Any tips / advice on how to load the images synchronously?

public/index.html
```diff
@@ -42,9 +42,6 @@
       Learn how to configure a non-root public URL by running `npm run build`.
     -->
     <title>SIE | Scientific Study Platform</title>
-    <!-- jsPsych library core script and styling -->
```
venjason, line 45: Decided to remove JsPsych library & stylesheets since it is not needed throughout application.

src/stories/components/ImageSelectionTask/ImageSelectionTask.js
```diff
@@ -17,21 +18,30 @@ import { Fragment } from 'react';
...
+      {/*  is most like you. To select the image on the left side press the */}
+      {/* E key;*/}
+      {/*  to select the image on the right press the I key.</p>*/}
+      {/* <p>This task is timed and will take approximately 20 minutes to */}
+      {/* complete.*/}
+      {/*  There are no wrong answers, so GO AS FAST AS YOU CAN.*/}
+      {/*  Responses that are too slow will not be recorded.</p>*/}
+      {/* <p>To get started, <a href={url}>click here</a>.</p>*/}
+
+      {/* TODO: Integrate JsPsych component into ImageSelectionTask to load on
```
venjason, line 41: Temporarily including JsPysch here inside ImageSelectionTask so that people can test the experiment. Will have to find a home for JsPsych experiment in future. 

Is the idea that we will be redirecting the user to a different url to take experiment OR are we utilizing ScreenTakeOver for the experiment?

src/stories/components/JsPsych/JsPsych.js
```diff
@@ -0,0 +1,240 @@
...
+  const [stimuliUrls, setStimuliUrls] = useState([]);
+  const [ready, setReady] = useState(false);
+
+  useEffect(() => {
+    // Load jsPsych library core script and styling & plugins used in the experiment
+    const jsPsychLibrary = document.createElement('script');
+    const jsPsychStyleSheet = document.createElement('link');
+    const htmlKeyboardResponsePlugin = document.createElement('script');
+    const imageKeyboardResponsePlugin = document.createElement('script');
+    jsPsychLibrary.src = '../../../external_libraries/jspsych-6.3.0/jspsych.js';
```
venjason, line 28: Potentially using the wrong path here... (discovered by Carlo)

Proper path should be: '../../../public/external_libraries/jspsych-6.3.0/jspsych.js';

With the current implementation, JsPsych runs perfectly fine for me in Chrome. Carlo was unable to get it to run with current path / updated path containing 'public'. Need to dig more into this. Please let me know if you are / unable to run experiment on your browser!

denglish7: Just tried to run this in Chrome and unfortunately it's not working for me either

src/stories/components/JsPsych/JsPsych.js
```diff
@@ -0,0 +1,240 @@
+/* eslint-disable */
+import React, { useEffect, useState } from 'react';
+import { useParams } from 'react-router-dom';
+import { observeStimuliCompletion } from '../../../firebase/api/gcp-utils';
+import { storeExperimentResult } from '../../../firebase/api/experiments';
+import example1 from '../../../images/example1.png';
+import example2 from '../../../images/example2.png';
```
fernandowinfield, line 7: Have you tried it like this: `import { example2 } from '../../../images/example2.png';`

src/stories/components/JsPsych/JsPsych.js
fernandowinfield: If the import syntax I suggested above doesn't work, then I would say let's just upload the example images to out cloud storage and pull them from there using HTTPs URLs.

src/stories/components/ImageSelectionTask/ImageSelectionTask.js
fernandowinfield: Let's use the ScreenTakeOver component that Alessia wrote. Let's try to have it be conditionally rendered using a ternary operator or something like that. Let's add a button in ImageSelectionTask component that opens the ScreenTakeover with all of the jsPsych stuff once it's clicked.

src/stories/components/JsPsych/JsPsych.js
fernandowinfield: I know we said this already but, have you tried again with loading the scripts and stylesheets (do we even need the stylesheets ever at all?) in `index.html`? To me that would be the method that's least likely to fail.

src/stories/components/ImageSelectionTask/ImageSelectionTask.js
venjason: Sounds good to me! Let's create a separate ticket to implement that feature in the next sprint? 

I tried playing around with the ScreenTakeOver previously to render the JsPysch component, but I believe with the constraints of Experiment.js, it only renders on the right hand side rather than the full screen.

src/stories/components/JsPsych/JsPsych.js
venjason: Sounds good to me! I believe that solution will work.

src/stories/components/JsPsych/JsPsych.js
venjason: I totally agree with you Fernando. The usual route that is taken when implementing JsPsych experiments is that the scripts & stylesheet would be loaded in index.html. 

However when I try to load it via index.html it does not properly load jsPsych library.

src/stories/components/JsPsych/JsPsych.js
venjason: (this is also the same behavior from previous teams implementation when they ran experiments)

src/stories/components/JsPsych/JsPsych.js
venjason: Just uploaded images to our cloud storage and images render properly now.

src/stories/components/JsPsych/JsPsych.js
```diff
@@ -0,0 +1,241 @@
...
+      stimulus: '<div style="height:307px; width:900px;"><div style="width:900px; height: 30px;  color:blue; margin-top:135px; font-size:60px;">+</div></div>',
+      choices: jsPsych.NO_KEYS,
+      trial_duration: 250, // From 1500
+      data: {label: 'post-fixation'},
+    };
+
+    // Call backend api storeExperimentResult to connect with FireBase and update Users Collection with experiment data.
+    async function saveExperimentData(experimentData) {
+      const newExperimentResult = {
+        experimentId: experimentData.get().json('pretty'),
```
venjason, line 181: FYI: Depending on the format of the data we want, this is where we are connecting with firebase to upload experimental data. If we want to filter out & pin point certain data points only, this is where we would do so.

src/stories/components/JsPsych/JsPsych.js
```diff
@@ -0,0 +1,241 @@
...
+      };
+      storeExperimentResult(participantId, experimentId, newExperimentResult);
```
venjason, line 183: FYI: Depending on the format of the data we want, this is where we are connecting with firebase to upload experimental data. If we want to filter out & pin point certain data points only, we would modify line 181.

denglish7 (commented):
I tried running this in chrome, but nothing happens when I get to the Image Selection Task section and try pressing any key to begin the task.

cmutuc25:
I think for dan's comment, we need to set the focus on the div that is receiving input events.

src/stories/components/JsPsych/JsPsych.js
fernandowinfield: Gotcha. I'm sure we could find a way to make them show up at the same time for an ideal user experience. But for now I think we can proceed as it is.

src/stories/components/ImageSelectionTask/ImageSelectionTask.js
fernandowinfield: I am ok with this 👍

fernandowinfield (approved):
@venjason, @denglish7, @cmutuc25,

Carlo's idea about the focus to receive input from the keyboard sounds right to me. But I think it would be more worth it to fix this once we have the screen takeover? For now, I can still run through the experiment if I click on the area where the instructions are shown (this probably shifts the focus like Carlo said).

In summary, I think this PR is ready to be merged assuming we'll do the following in separate PR(s):
- Add screen takeover and open it after participant clicks on a button that will show up in the ImageSelectionTask component (step 5 of the wizard).
- Fix the focus issue that Dan pointed out. Maybe this is no longer needed once we add the screen takeover as this will cover the entire screen?
- Organize the resulting data (following the format suggested by backend teams) and upload it to Firestore.
- Lastly, enable/disable the 'Next' button in step 5 of the wizard once the whole process is done.

How does this sound?

src/stories/components/JsPsych/JsPsych.js
venjason: Carlo actually has found a solution to this. Just waiting for this PR to be approved & changes for this will be coming. He added 'jspsych-preload` plugin which helped load the images so both images pop up at the same time!

src/stories/components/ImageSelectionTask/ImageSelectionTask.js
venjason: Note: Carlo has taken the responsibility for rendering JsPsych via ImageSelectionTask.js. There will be a separate task / PR for this feature.

src/stories/components/JsPsych/JsPsych.js
venjason: Yes! Thank you.

venjason:
Following up regarding Fernando's comment above.

That sounds like a great plan. Without getting this PR merged it is a huge roadblock for us being more productive. 

Here are the remaining tasks that need to be implemented for JsPsych (separate tasks / PR's):

1. Integrating ImageSelectionTask.js with JsPsych (potentially using ScreenTakeOver.js) - this task is being handled by Carlo
2. Transforming the resulting data to follow the back-end teams scheme & upload it to Firestore - this task is being handled by me (Jason)
3.  Enable/Disable the 'next' button in step 5 once whole experiment is done - not assigned. I believe this task can go hand in hand with the first task that was mentioned. Will forward this information to Carlo to see if he is comfortable with this.

PR #89, opened 2021-04-08
[SIE 141] Consent form integration

fernandowinfield:
Integrated Consent form Qualtrics survey into Experiment flow. Had to add a couple of parameters to the Wizard component to be able to communicate state between that component and the Experiment component.

To test:
- `npm start`
- Go to http://localhost:3000/9iG2wajKXkjLXiLQfwgj/36qwRoyuSzcHzvCAapya4hrJBgQ2
- Move to step 2 of the Wizard and complete the Qualtrics survey. If agreed to the consent form, the 'Next' button should be enabled and you should be able to continue to the next steps of the experiment.
- Note: these steps assume that the user doc `36qwRoyuSzcHzvCAapya4hrJBgQ2` does not have any Experiment collections in Firestore (in other words, the participant must not have done this experiment before).

Jira [ticket](https://cs6510.atlassian.net/browse/SIE-141)

### DIFF
--- a/src/stories/layouts/Wizard/Wizard.js
+++ b/src/stories/layouts/Wizard/Wizard.js
@@ -12,8 +12,10 @@ import { Link, useParams } from 'react-router-dom';
  * Component for Wizard layout element.
  * @param {node} children of the component
  * @param {array} labels of steps
+ * @param {bool} showNext Whether to enable the 'Next' button or not
+ * @param {func} stepHandler React hook to share the current step of the Wizard
  * @return {object} (
- *   <Wizard labels={labels}>
+ *   <Wizard labels={labels} showNext={showNext} stepHandler={stepHandler}>
  *       {children}
  *   <Wizard />
  * )
@@ -103,7 +105,10 @@ const Controls = ({ experimentId, currentStep, WizInstance, showNext }) => (
           <Link className={'wizard__button button button--tertiary'}
             to={`/study/${experimentId}/user/123`}
           >Complete Study</Link> :
-          <button className={'wizard__button button button--tertiary disabled'}
+          <button className={'wizard__button ' +
+            'button ' +
+            'button--tertiary ' +
+            'button--disabled'}
           >Next</button>
       }
     </div>
--- a/src/stories/layouts/Wizard/Wizard.js
+++ b/src/stories/layouts/Wizard/Wizard.js
@@ -19,7 +19,7 @@ import { Link, useParams } from 'react-router-dom';
  * )
  */
 /* eslint react/prop-types: 0 */
-export const Wizard = ({ children, labels, showNext }) => {
+export const Wizard = ({ children, labels, showNext, stepHandler }) => {
   const { experimentId } = useParams();
   const [state, updateState] = useState({
     form: {},
@@ -29,6 +29,7 @@ export const Wizard = ({ children, labels, showNext }) => {
   // update step value
   const onStepChange = (stats) => {
     setStep(WizInstance.currentStep);
+    stepHandler(WizInstance.currentStep);
     window.setTimeout(function() {
       window.scrollTo({
         top: 0,
@@ -119,9 +120,13 @@ Wizard.propTypes = {
      */
   labels: PropTypes.array,
   /**
-     * Wizard's 'Next' button condition
+     * Wizard's 'Next' button condition (whether to show it or not)
      */
   showNext: PropTypes.bool,
+  /**
+     * React hook to share the current step of the Wizard
+     */
+  stepHandler: PropTypes.func,
 };
 
 Wizard.defaultProps = {
--- a/src/stories/layouts/Wizard/Wizard.js
+++ b/src/stories/layouts/Wizard/Wizard.js
@@ -142,4 +142,6 @@ Wizard.defaultProps = {
     'Step 3',
     'Step 4',
   ],
+  showNext: true,
+  stepHandler: (step) => step,
 };
--- a/src/stories/layouts/Wizard/Wizard.js
+++ b/src/stories/layouts/Wizard/Wizard.js
@@ -98,18 +98,13 @@ const Controls = ({ experimentId, currentStep, WizInstance, showNext }) => (
           onClick={WizInstance.previousStep}
         >Go Back</button>
       }
-      {(currentStep !== WizInstance.totalSteps && showNext) ?
+      {currentStep === WizInstance.totalSteps ?
+        <Link className={'wizard__button button button--tertiary'}
+          to={`/study/${experimentId}/user/123`}
+        >Complete Study</Link> :
         <button className={'wizard__button button button--tertiary'}
-          onClick={WizInstance.nextStep}
-        >Next</button> : currentStep === WizInstance.totalSteps ?
-          <Link className={'wizard__button button button--tertiary'}
-            to={`/study/${experimentId}/user/123`}
-          >Complete Study</Link> :
-          <button className={'wizard__button ' +
-            'button ' +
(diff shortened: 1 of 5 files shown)


src/stories/layouts/Wizard/Wizard.stories.js
```diff
@@ -13,36 +14,38 @@ export default {
...
-      </div>
-      <div className="step-4">
-        <h3>Step 4</h3>
-                Lorem ipsum dolor sit amet, an has summo
-                riure epicuri, has illud rationibus et. Prima ridens sit te,
-                nam idque explicari expetendis in. An mei adolescens mnesarchum,
-                ei periculis adipiscing per, probo populo nec ad.
-      </div>
-    </Wizard>
+    <MemoryRouter initialEntries={['/user/Pl3WJYa7vQ1ALVt0rHRV/study/108']}>
```
harmanwho, line 17: I am not sure what these hardcoded values mean, but the rest of the code looks good to me. I pulled the branch and checked the steps, it seems to be working great.

src/stories/layouts/Wizard/Wizard.js
```diff
@@ -29,6 +31,7 @@ export const Wizard = ({ children, labels }) => {
   // update step value
   const onStepChange = (stats) => {
     setStep(WizInstance.currentStep);
+    stepHandler(WizInstance.currentStep);
```
harmanwho, line 34: This seems redundant as `stepHandler` and `setStep` do basically the same thing but in parent and child component. I tried to remove `setStep` and use stepHandler and its step value but Next button disables as it doesn't fetch the latest step.

harmanwho: or maybe we can just remove the `step` altogether and yes just `WizInstance.currentStep` wherever needed; for example, while passing it in the Controls component.
`{
          WizInstance && <Controls currentStep={step -> WizInstance.currentStep}
            WizInstance={WizInstance}
            experimentId={experimentId}
            showNext={showNext} />
        }`

fernandowinfield: I think the difficulty comes from having to know what the is going on with the Wizard but in the Experiment component. So `stepHandler` helps to tell Experiment what step the Wizard is in. `setStep` is only used by the Wizard to update the step. So, not sure if can live with only one of the two.

If there is a way to access WizInstance's functionality in Experiment then I think we could try what you suggest in your second comment. Do you know how to access that from within Experiment?

src/stories/layouts/Wizard/Wizard.stories.js
fernandowinfield: These are just values that mock the URL parameters (`experimentId` and `participantId`). We need this in the Storybook story to avoid crashing the app when we use `useParams()`

src/stories/layouts/Wizard/Wizard.js
harmanwho: So I replaced the `currentStep` attribute of `<Control>` with `WizInstance.currentStep` and it works perfectly fine
Screen Shot 2021-04-08 at 1 53 44 PM

harmanwho: Also, 

> I think the difficulty comes from having to know what step the Wizard is in but in the Experiment component. So `stepHandler` helps to tell Experiment what step the Wizard is in. `setStep` is only used by the Wizard to move back and forth between steps. So, not sure if we can live with only one of the two.
> 
> If there is a way to access WizInstance's functionality in Experiment then I think we could try what you suggest in your second comment. Do you know how to access that from within Experiment?

`stepHandler` stays because it is used in experiment.js but the one we can remove is the local `step` state for Wizard.js
Screen Shot 2021-04-08 at 1 58 52 PM

harmanwho: It's not a big deal, it is just one more state haha

alecherryy:
@harmanwho and @fernandowinfield is this PR good to go?

src/stories/layouts/Wizard/Wizard.js
```diff
@@ -85,21 +90,26 @@ export const Wizard = ({ children, labels }) => {
...
-        >Completet Study</Link>
+        >Next</button> : currentStep === WizInstance.totalSteps ?
+          <Link className={'wizard__button button button--tertiary'}
+            to={`/study/${experimentId}/user/123`}
+          >Complete Study</Link> :
+          <button className={'wizard__button ' +
+            'button ' +
+            'button--tertiary ' +
+            'button--disabled'}
+          >Next</button>
```
alecherryy, line 112: I would change this to this:
```
{currentStep === WizInstance.totalSteps ?
        <Link className={'wizard__button button button--tertiary'}
          to={`/study/${experimentId}/user/123`}
        >Complete Study</Link> :
        <button className={'wizard__button button button--tertiary'}
          onClick={WizInstance.nextStep} disabled={!showNext}
        >Next</button>
      }
```

You can use the disabled attribute to prevent the button from working and not needing extra logic.

alecherryy (changes requested):
@fernandowinfield I run this locally and I am able to go to the next steps in the wizard without completing the consent form.

src/stories/layouts/Wizard/Wizard.js
fernandowinfield: Gotcha, that makes sense 👍. I didn't write the Wizard component and it doesn't have to do with what this PR is trying to accomplish so I think we can remove the `step` variable in a separate PR to keep things well contained?

fernandowinfield:
@alecherryy

> I run this locally and I am able to go to the next steps in the wizard without completing the consent form.

I think the reason is because of the 4th bullet point I wrote for the description of this PR:

"
- Note: these steps assume that the user doc 36qwRoyuSzcHzvCAapya4hrJBgQ2 does not have any Experiment collections in Firestore (in other words, the participant must not have done this experiment before).

"

I used your suggestion about the Wizard controls next button logic and I think it is a lot cleaner. But I decided to still include some styling to make the disabled button grey. I felt like this is better to convey that they should complete the Qualtrics survey before the 'Next' button gets enabled. But if that doesn't make sense I can get rid of such styling.

alecherryy (changes requested):

alecherryy:
> @alecherryy
> 
> > I run this locally and I am able to go to the next steps in the wizard without completing the consent form.
> 
> I think the reason is because of the 4th bullet point I wrote for the description of this PR:
> 
> "
> 
>     * Note: these steps assume that the user doc 36qwRoyuSzcHzvCAapya4hrJBgQ2 does not have any Experiment collections in Firestore (in other words, the participant must not have done this experiment before).
> 
> 
> "
> 
> I've reset the user doc so you can test it again; this time it should work.
> 
> Also, I used your suggestion about the Wizard controls next button logic and I think it is a lot cleaner. But I decided to still include some styling to make the disabled button grey. I felt like this is better to convey that they should complete the Qualtrics survey before the 'Next' button gets enabled. But if that doesn't make sense I can get rid of such styling.

Yes, good call on the styling! I didn't mention it, but that's perfect!

fernandowinfield:
@alecherryy I think this last change is closer to what you had mentioned about using the disabled attribute rather than a class. Per your suggestion (thanks, makes a lot of sense), I moved the disabled styling of the button from the Wizard's styles into the Button's styles. The only thing is I had to use `!important` to override the teal `background-color` of the `.wizard__button.button--tertiary` class. I hope that's a sparingly enough use of `!important` but if you see a better way to do it let me know

alecherryy:
LGTM!

alecherryy (approved):

PR #97, opened 2021-04-19
Experiment task full screen

cmutuc25:
Used the `Modal` component to have the jsPsych selection task take up the whole screen. `ScreenTakeover` was giving me issues with styling.

- There's a button that opens up the `Modal` and shows the selection task on Step 5
- `Next` button to go to the next step is disabled until the experiment is complete

Some ui/ux issues I found that I could use some input on, we can put these into a new ticket but wanna get some ideas first:

1. Once the modal is open, the focus isn't on the selection div so we need to fix that again. You have to click on the "task" for key input events to work.
2. After the experiment ends, you need to press a key to fully complete and got to the next step. Also once you complete, the data gets input onto the screen?
3. Is you close the modal before the selection task ends, it still runs in the background, does jsPsych have a way to pause experiments?

To test
- npm start
- [http://localhost:3000/9iG2wajKXkjLXiLQfwgj/Bo0Hka0QT5O2Q026qKpFbAYHOW23/](http://localhost:3000/9iG2wajKXkjLXiLQfwgj/Bo0Hka0QT5O2Q026qKpFbAYHOW23/)
- go to step 5
- notice that next button is disabled
- press button to open modal
- run the experiment all the way through
- check if the button is enabled


venjason:
Regarding the data displaying after the experiment concludes, there's one line of code you can remove. Should look something similar to this below in JsPsych.js:
`

              TODO: can remove displayData() once we find a home for JsPsych

              jsPsych.data.displayData();
`
If you comment / delete this line out, it will no longer display the data on click. 

I just tested it out and can confirm that the button gets enabled upon experiment completion.

venjason:
This is something we should consider:

- The experiment does not fully finish (does not store / save result) until the user clicks another key after the last 'timeline trial' which would be the `const reportBlock`. 
   - This is the screen that shows the user the number of trials / stats about selections. 
   - Once any button is clicked, it will activate 'on_finish' tasks. Example shown below:
```
jsPsych.init({
          timeline: timeline,
          display_element: 'jspsych-target',
          on_finish: function () {
            saveExperimentData(jsPsych.data);
          },
        });
```
- If we aren't able to figure out how to close the pop-up window on a click that would also save the experiment at the same time, a work-around would be storing / saving the data in the 'reportBlock' timeline. The function would be called there instead, and functionally would still work the same since the data has already been generated.

venjason:
Another behavior I noticed is that once the experiment has been completed, the user still has access to the button to open the modal once again. (referring to the button '`Start Experiment`')

src/stories/components/JsPsych/JsPsych.js
```diff
@@ -81,44 +80,50 @@ export const JsPsych = () => {
         };
         timeline.push(instructions);
 
+        // Preload images for experiment
+        var preload = {
```
venjason, line 84: Preload plug in works amazing!! This is fabulous that we are able to render both images at the same time now.

venjason:
I believe there is a way to pause the experiment with built in JsPsych functions. Please refer to https://www.jspsych.org/core_library/jspsych-core/#jspsychpauseexperiment 

Example I found:
`
```
jsPsych.pauseExperiment();

jsPsych.resumeExperiment();
```
`

venjason (approved):
Follow-up per our meeting this morning, the plan is to split this task into smaller PRs. 

Info regarding this feature:
- Confirmed that a button 'Start Experiment' appears on Step 5 (selection task) and the 'Next' button is disabled. 
- When user clicks on 'Start Experiment', Modal appears with JsPsych experiment ready to be conducted
- When user finishes JsPsych experiment, the user has to close out of the Modal & the 'Next' button is enabled for them to proceed to Step 6. 
- Behavior of images not loading at the same time has been fixed via preload plugin. Images load simultaneously now.

PR #98, opened 2021-04-19
Updating saveExperimentData method to refractor data & upload to firebase via new backend API

venjason:
Creating a method to transform result data from JsPsych experiment to match model scheme of back end api. Utilizes new backend API 'uploadSelectionResult' via gcp-utils.js 

Confirmed that data gets properly stored in FireBase bucket 'sie-user-selections' as 'result.csv' for this particular user / experiment (Bo0Hka0QT5O2Q026qKpFbAYHOW23-9iG2wajKXkjLXiLQfwgj).

Here are some notes regarding the input / output for the trial experiment: 
- If user makes the 'correct' choice - meaning they chose the original image - it gets counted as a '1' in response column.
- If user makes the 'different' choice - meaning they chose the inverted image - it gets counted as a '-1' in response column.
- If user makes no selection, it gets counted as '-1'.

How I tested:
- Check current data in firebase bucket via https://console.firebase.google.com/u/0/project/cs6510-spr2021/storage/sie-user-selections/files~2FBo0Hka0QT5O2Q026qKpFbAYHOW23-9iG2wajKXkjLXiLQfwgj 
- npm start
- http://localhost:3000/9iG2wajKXkjLXiLQfwgj/Bo0Hka0QT5O2Q026qKpFbAYHOW23
- go to step 5
    - option: modify 'setPhotoStepCompleted' to 'true' to bypass image upload
- Finish the study with a different set of answers than data that is in firebase
- Confirm that the data gets updated in firebase bucket

### DIFF
--- a/src/stories/components/JsPsych/JsPsych.js
+++ b/src/stories/components/JsPsych/JsPsych.js
@@ -1,8 +1,10 @@
 /* eslint-disable */
 import React, { useEffect, useState } from 'react';
 import { useParams } from 'react-router-dom';
-import { observeStimuliCompletion } from '../../../firebase/api/gcp-utils';
-import { storeExperimentResult } from '../../../firebase/api/experiments';
+import {
+  observeStimuliCompletion,
+  uploadSelectionResult
+} from '../../../firebase/api/gcp-utils';
 
 /**
  * Component to load jsPsych's "Self Image Experiment".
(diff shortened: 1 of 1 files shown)


theodore3131 (approved):
LGTM! Thanks for the quick update! One thing I wonder if the frontend need to the catch the response from uploadSelection() to tell the user that it's safe to move to next step?

fernandowinfield:
@venjason, I think what @theodore3131 is saying makes sense and would be great for us to be able to know when we can close the modal/screen takeover and enable the wizard's 'Next' button.

I'll leave it up to you if you want to add that here, but because it might not be as trivial, I think it also makes sense to do that part in a separate PR. I can give some ideas on how to achieve that if you'd like.

PR #100, opened 2021-04-20
added feedback messages to signup

denglish7:
Added feedback messages for signup to react to user mistakes

https://cs6510.atlassian.net/secure/RapidBoard.jspa?rapidView=4&projectKey=SIE&view=planning&selectedIssue=SIE-153&issueLimit=100

### DIFF
--- a/src/stories/pages/Signup/Signup.js
+++ b/src/stories/pages/Signup/Signup.js
@@ -40,32 +40,54 @@ export const Signup = ({ isDarkTheme }) => {
     'Researcher',
   ];
 
+  const validate = () => {
+    setError('');
+    let isError = false;
+
+    if (firstName === '') {
+      setError('Please enter a first name');
+      isError = true;
+    } else if (lastName === '') {
+      setError('Please enter a last name');
+      isError = true;
+    } else if (role === '') {
+      setError('Please select a role');
+      isError = true;
+    }
+
+    return isError;
+  };
+
   const postUser = (e) => {
     e.preventDefault();
     // call user service
     // define new user object
-    const user = {
-      firstName,
-      lastName,
-      email,
-      password,
-      role,
-    };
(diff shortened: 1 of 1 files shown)


src/stories/pages/Signup/Signup.js
```diff
@@ -40,32 +40,54 @@ export const Signup = ({ isDarkTheme }) => {
...
+    setError('');
+    let isError = false;
+
+    if (firstName === '') {
+      setError('Please enter a first name');
+      isError = true;
+    } else if (lastName === '') {
+      setError('Please enter a last name');
+      isError = true;
+    } else if (role === '') {
```
stenizwahyudi, line 53: I might insert other conditional statements for empty email and password. I know that the email and password will be checked by the backend, but I think it would be nicer if we give error messages when the fields are empty. Please consider the case below:
**Assuming the first name and last name fields are filled out, and email, password, role are left blank, I would expect to be notified that the email is empty, but the user is notified that the role field is empty as of now because the checking process for email and password hasn't been done in this case.

    } else if (email === '') {
      setError('Please enter a valid email address');
      isError = true;
    } else if (password === '') {
      setError('Please enter a password that is 6 characters long or more');
      isError = true;

Other than that, I think the pr looks good. Unless there are any other feedbacks.

stenizwahyudi (changes requested):

src/stories/pages/Signup/Signup.js
denglish7: I think you're absolutely right. No point in making a call to the backend to check if it's a valid email if there's no characters anyways. I'll make the changes. Thanks!

venjason: Updated changes look good 👍

stenizwahyudi (approved):

PR #101, opened 2021-04-20
[SIE-185] Fix Next button incorrectly enabling

fernandowinfield:
TLDR: fixed two bugs: (1) make sure that the 'Next' button stays disabled when face detection fails and (2) make sure we the 'Next' button is disabled again if the participant submits an invalid photo after a valid one. Kudos to @stenizwahyudi  for finding (1) and @quanganhhoang for (2).

-----

See [this Jira ticket](https://cs6510.atlassian.net/browse/SIE-185) for more info on reproducing bug # 1. My guess is that the `getConsentForm()` listener stays open even after moving on to the next steps of the wizard and mistakenly updates `showNext`, which is the state boolean that controls the enabling/disabling of the 'Next' button.

In this PR:
- Bug # 1: Slightly modified `getConsentForm()` to close the `onSnapshot()` listener once it successfully returns. @bleachalon and/or @theodore3131 , it would be great if you could give this a quick look to make sure I am not making a really bad mistake.
- Bug # 2: Modified the logic in UploadPhoto and Experiment components to make sure that 'Next' button becomes disabled in the case that a photo had passed face detection but the most recent photo didn't pass. @quanganhhoang and FE peeps if you could take a look to double check my logic for this part that would be great.


stenizwahyudi:
The disabling and enabling of the Next button are behaving correctly, but the Travis build is currently failing.

src/stories/pages/Experiment/Experiment.js
```diff
@@ -82,6 +80,20 @@ export const Experiment = ({
...
+      setShowNext(false);
+      setCompletedSteps((prevState) => {
+        // Remove `3` from the completed steps array
+        const prevCompletedSteps = [...prevState];
+        const index = prevCompletedSteps.indexOf(3);
+        if (index > -1) {
+          prevCompletedSteps.splice(index, 1);
+        }
+        return prevCompletedSteps;
+      });
```
cmutuc25, line 96: This should be fine as long as `photoStepCompleted` is set to `false` correctly whenever we need to get into the `else if`.

cmutuc25 (approved):

PR #104, opened 2021-04-20
Updating Modal/Button class to contain an additional parameter to disable button

venjason:
We want to hide/disable "Start Experiment" button (in image selection task section) when the task is finished. Reason being we don't want users to be able to reopen modal. We'd rather disable the button, so that they're only option is to click the "Next" button to continue with the experiment sections. 

Here are the major changes:

- Updating Button.js to contain an additional parameter "disableBtn" which will be used to set the react button class's 'disabled' field. If the user finishes the selection task, the button will be disabled. 
- Updating Modal.js to reflect changes mentioned above - adding a 'disabledBtn' class to the Button component.
- Changing 'canClose' variable name to 'selctionTaskComplete' so that we can use it for both Modal close feature + disable button feature.

Expected behavior:
- Users can click "Start Experiment" to open up modal which loads JsPsych component to conduct image selection task. 
- When users are done with JsPsych component, they are able to close out of the modal, and they will see that the "Start Experiment" button is disabled / hidden.

### DIFF
--- a/src/stories/components/Button/Button.js
+++ b/src/stories/components/Button/Button.js
@@ -19,7 +19,8 @@ import { Link } from 'react-router-dom';
  * )
  */
 export const Button =
-({ modifierClasses, url, text, isButton, onClick }) => {
+({ modifierClasses, url, text, isButton,
+  onClick, disableBtn }) => {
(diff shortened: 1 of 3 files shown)


cmutuc25 (approved):
looks good to me, out of date btw

PR #106, opened 2021-04-21
Fix Modal open and close

cmutuc25:
Added `canOpen` and `canClose` to Modal to fix issue where sometimes you can't open and close a Modal if the shared prop wasn't set.


venjason (approved):
Works as expected! Thanks for catching the bug Carlo. LGTM.

PR #110, opened 2021-04-22
Anonymous auth for Experiments

cmutuc25:
Implemented automatic aunonymous auth login when accessing the Experiments component.

This auth state is needed to call endpoints from firebase.

If logged in anonymously, cannot access any other routes.

### DIFF
--- a/src/App.js
+++ b/src/App.js
@@ -65,15 +65,19 @@ function App() {
             </Dashboard>
           )} />
         <Route exact path="/login">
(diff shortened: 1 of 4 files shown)


fernandowinfield:
I am getting an `undefined` error when trying to access `.email` of `user`. Maybe just use `?.` in those cases?

fernandowinfield (approved):

